PR #39, opened 2022-09-30
feat(cloudnative-pg): add user facing roles view and edit

NaPs:
This PR implements the k8s default user facing roles (view, edit, admin). More infos here: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

### DIFF
--- a/charts/cloudnative-pg/README.md
+++ b/charts/cloudnative-pg/README.md
@@ -44,6 +44,7 @@ CloudNativePG Helm Chart
 | podLabels | object | `{}` | Labels to be added to the pod. |
 | podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security Context for the whole pod. |
 | priorityClassName | string | `""` | Priority indicates the importance of a Pod relative to other Pods. |
+| rbac.aggregateClusterRoles | bool | `true` | Aggregate ClusterRoles to Kubernetes default user-facing roles. Ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles |
 | rbac.create | bool | `true` | Specifies whether ClusterRole and ClusterRoleBinding should be created. |
 | replicaCount | int | `1` |  |
 | resources | object | `{}` |  |
--- a/charts/cloudnative-pg/README.md
+++ b/charts/cloudnative-pg/README.md
@@ -44,7 +44,7 @@ CloudNativePG Helm Chart
 | podLabels | object | `{}` | Labels to be added to the pod. |
 | podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security Context for the whole pod. |
 | priorityClassName | string | `""` | Priority indicates the importance of a Pod relative to other Pods. |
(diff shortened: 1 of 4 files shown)


charts/cloudnative-pg/templates/rbac.yaml
```diff
@@ -378,4 +378,54 @@ subjects:
 - kind: ServiceAccount
   name: {{ include "cloudnative-pg.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
phisco, line 382: I'd guard the creation of these ClusterRoles with a variable, either using `rbac.aggregateClusterRoles` or adding a second flag, defaulting to `false`. So that people upgrading to the new version won't grant additional permissions to their users by mistake.

phisco (changes requested):

charts/cloudnative-pg/templates/rbac.yaml
NaPs: (Cluster)Roles do not grant any permissions without a matching (Cluster)RoleBinding. Maybe defining `rbac.aggregateClusterRoles` to false by default would be enough to achieve your requirement without needing an extra values variable?

phisco: Sounds good to me, if you can change that, @NaPs :+1:  I'd say this could also be backported to the plain manifest installation method, I'll open an issue for that.

charts/cloudnative-pg/values.yaml
```diff
@@ -67,6 +67,9 @@ serviceAccount:
 rbac:
   # -- Specifies whether ClusterRole and ClusterRoleBinding should be created
   create: true
+  # -- Aggregate ClusterRoles to Kubernetes default user-facing roles.
+  # Ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles
+  aggregateClusterRoles: true
```
phisco, line 72:
```suggestion
  aggregateClusterRoles: false
```

charts/cloudnative-pg/README.md
```diff
@@ -39,6 +39,7 @@ CloudNativePG Helm Chart
 | nodeSelector | object | `{}` | Nodeselector for the operator to be installed |
 | podAnnotations | object | `{}` | Annotations to be added to the pod |
 | podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security Context for the whole pod |
+| rbac.aggregateClusterRoles | bool | `true` | Aggregate ClusterRoles to Kubernetes default user-facing roles. Ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles |
```
phisco, line 42:
```suggestion
| rbac.aggregateClusterRoles | bool | `false` | Aggregate ClusterRoles to Kubernetes default user-facing roles. Ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles |
```

phisco (changes requested):

sathieu:
@NaPs  What is the status of this PR?

NaPs:
Hi @sathieu, my bad I completely left this issue behind... I'll do the suggested modifications and rebase the branch.

NaPs:
@phisco All done :-)

phisco (approved):

PR #48, opened 2022-10-14
Updates sandbox monitoring configmap to cnpg-default-monitoring

jacobfederer:
As described here: https://github.com/cloudnative-pg/cloudnative-pg/issues/811 I ran into an issue with deploying an example cluster with the sandbox. The cluster could not find the 'cnpg-default-monitoring' config map since it wasn't copied to the custom namespace. This might be due to a misconfiguration of the sandbox default values.

### DIFF
--- a/charts/cnpg-sandbox/values.yaml
+++ b/charts/cnpg-sandbox/values.yaml
@@ -74,7 +74,7 @@ cloudnative-pg:
   config:
     create: true
     data:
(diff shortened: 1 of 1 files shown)


phisco:
we can directly avoid setting the env variable regarding monitoring queries as it's already handled by default by the cloudnative-pg chart

phisco:
Will be addressed in #52, thanks @jacobfederer for your heads-up and time spent debugging it, I'll put you as a co-author!

PR #82, opened 2023-02-20
fix: use wget to check service availability

jakubhajek:
The busybox image does not contain the `curl` command to perform tests. Thus, the pod failed to start with the following error message:

```sh
unable to start container process: exec: "curl": executable file not found in $PATH: unknown
```
This PR replaces the curl command with `wget` and updates the argument to test the service availability.


jakubhajek:
However, I do not completely understand what is the purpose of this test. The endpoint returns 404, at least this is the response I always get in my test environments. It causes that after upgrading the CNPG controller we have the test-connection pod with failed status which has to be manually deleted.

phisco:
Yes, it's not really testing anything we should just remove it.

jakubhajek:
Thanks, Philippe for confirming this. That being said we can close this PR.

PR #88, opened 2023-03-20
Proposed values.yaml for a cluster chart + some basic chart setup

itay-grudev:
The key controversial point is the `mode` parameter I propose that would define an entirely different initialization behaviour of the cluster.

My proposal is the following modes of operation:

* `standalone`
* `replica_cluster` - https://cloudnative-pg.io/documentation/current/replica_cluster/
* `recovery` - https://cloudnative-pg.io/documentation/current/bootstrap/#bootstrap-from-a-backup-recovery
* `recovery_external` - https://cloudnative-pg.io/documentation/current/bootstrap/#recovery-from-an-object-store
* `recovery_pg_basebackup` - https://cloudnative-pg.io/documentation/current/bootstrap/#bootstrap-from-a-live-cluster-pg_basebackup

I propose we review it and make changes. If we agree on a version of it, we could use it as a starting point for the remainder of the chart.


gbartolini:
IMHO we need to differentiate use cases between replica cluster and recovery (full or PITR). I would therefore consider these two main possibilities:

- replica cluster (meaning that replica must be enabled):
   - via recovery object store, with optional streaming async replication
   - streaming only (using pg_basebackup and streaming replication)
- recovery from an object store:
   - full
   - PITR (target specification)

Ideally, in the replica cluster we should be able to provide the full topology of the distributed cluster, meaning that the external servers component should be the same on all clusters.

For the replica cluster, I would rather start with the bootstrap from the recovery object store - I honestly do not see practical use cases having a replica cluster without backups configured.

As a result, we can come up with 3 use cases:

- standalone
- replica cluster (from object store, with streaming - optional)
- recovery (full or PITR, based on target specification)

Sounds reasonable?

itay-grudev:
I just pushed an alternative based on your feedback.

```yaml
##
# Cluster mode of operation. Available modes:
# * `standalone` - default mode. Creates new or updates an existing CNPG cluster.
# * `replica` - Creates a replica cluster from an existing CNPG cluster. # TODO
# * `recovery` - Same as standalone but creates a cluster from a backup, object store or via pg_basebackup. # TODO
mode: standalone

##
# Recovery settings if the chosen mode is `recovery`.
recovery:
  ##
  # Available recovery methods:
  # * `backup` - Recovers a CNPG cluster from a CNPG backup (PITR supported) Needs to be on the same cluster in the same namespace. # TODO
  # * `object_store` - Recovers a CNPG cluster from a barman object store (PITR supported). # TODO
  # * `pg_basebackup` - Recovers a CNPG cluster viaa streaming replication protocol. Useful if you want to
  #        migrate databases to CloudNativePG, even from outside Kubernetes. # TODO
  method: backup
```

itay-grudev:
Closed in favor of #106

PR #98, opened 2023-04-24
extraDeploy value to add arbitrary object to the release

jouve:
the allows users of the chart to add arbitrary object  in the context of the chart : to use templates like `cloudnative-pg.selectorLabels` and values for consistency), and to manage them with the same workflow as the rest (helm upgrade, uninstall, ...)

example of resources are podmonitor or networkpolicy.


phisco:
I'm not sure I like the approach, depending on the tool you are using you could achieve the same thing in many other ways 🤔   You could even define your own umbrella Chart, having CNPG as a dependency and defining any additional resource you want as template

sxd:
@jouve my understanding here it's the idea of using this chart as a meta package to install other stuff using the same helm chart right? The approach of the operator it's to be a single solution and if you need something else to be deployed along side the operator, you should have an external tool to manage that not just one chart.

In my opinion, if you need a field for "extra objects" the approach isn't the right one since that should be managed by an external tool, for me this it's not a good idea or at least something we should handle in the chart.

I'm tempted to close this PR and probably we can discuss this on slack, what do you think @phisco @jouve  ?

Regards

jouve:
the idea is not to deploy random stuff, but related & coupled resources :)

as in the example , the pod monitor is coupled by the `cloudnative-pg.selectorLabels`  and does make sense only deployed with the chart.

Also this PR does not incur any maintenance for you once merged.

Ornias1993 (changes requested):
"extraDeploy" is confusing with "deployments".

Ornias1993:
> the idea is not to deploy random stuff, but related & coupled resources :)
> 
> as in the example , the pod monitor is coupled by the `cloudnative-pg.selectorLabels` and does make sense only deployed with the chart.
> 
> Also this PR does not incur any maintenance for you once merged.

@jouve If you want this, we just merged a advanced variant of this to our TrueCharts common library.
Our CNPG chart should come out later this week, including this feature.

Ornias1993:
@jouve It's finished and tested by now, I think this should give you the option to use an alternative helm chart with more broad configuration options available. See: https://truecharts.org/charts/operators/cloudnative-pg/

as it's based on our common-chart, all options and more defined here are available: 
- https://truecharts.org/manual/helm/common/
- https://github.com/truecharts/library-charts/blob/main/library/common/values.yaml

phisco:
Yeah, I don't think we are going to merge this. Thanks @Ornias1993 for proposing your chart, it's awesome to see more options available from the community. 
On the other hand, personally, I would argue that approach is actually inverting the dependency. The way I see it is that charts should give you the knobs to customize the app being deployed in such a way that it would work with any other resource you defined externally. For example, in the podmonitor use case above, you could define a chart or just use a plain manifest to be deployed alongside the our chart, which could use the labels you configured via the values.yaml file of the cnpg chart.

Ornias1993:
> 

I agree the usecase listed in this PR is not really making sense in a helm/kube way.

For our Usecase as TrueCharts, the usecase is different... We also build for GUI based kubernetes control software, which does not always easily offer the option to add manifests. We also explicitly wanted to expose access to our common includes, which cannot be done by creating external manifest files.

But yes, in TLDR: Same idea, different goals than this PR.

PR #102, opened 2023-04-26
Drop obsolete test pod values

jcpunk:
My auditors want fully qualified container images, I figured others would probably benefit from this as well.


phisco (approved):

phisco:
Oops, we actually removed the test pod some time ago, you can directly remove the whole test section from the values!

jcpunk:
Updated to just drop the test values

phisco (approved):

PR #103, opened 2023-04-27
feat: add metrics from cluster-example-monitoring.yaml

jcpunk:
This adds the metrics from cloudnative-pg/blob/docs/src/samples/cluster-example-monitoring.yaml

Since these metrics are in the sample, it is nice to have them in the defaults.

### DIFF
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -510,3 +510,178 @@ monitoringQueriesConfigMap:
         - setting:
             usage: "GAUGE"
             description: "Setting value"
+    pg_stat_activity:
+      query: |
+        WITH
+          metrics AS (
+            SELECT
+              application_name,
+              SUM(EXTRACT(EPOCH FROM (CURRENT_TIMESTAMP - state_change))::bigint)::float AS process_idle_seconds_sum,
+              COUNT(*) AS process_idle_seconds_count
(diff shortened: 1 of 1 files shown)


phisco (approved):

jcpunk:
I think this has all the same metrics as the sample files, but my by hand reconcile may have missed something...

phisco:
ok, now i got it, it's in sync with the default ones we deploy with manifests, see https://github.com/cloudnative-pg/cloudnative-pg/blob/main/releases/cnpg-1.20.0.yaml#L12160. The ones you are adding come from the example one: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/cluster-example-monitoring.yaml#L27. I'm not sure it was a deliberate choice not to include those by default tbh, I'll have to check!

jcpunk:
Interesting, I didn't realize they were also in the release yaml. I'd love for all 3 to have the same metrics.  This way I've got a consistent metrics set in whichever mode I've loaded them from.

itay-grudev:
@litaocdl @jcpunk @phisco I am confused about the state of this PR. It predates my work on `cloudnative-pg/charts`. I thought it was outdated, but I noticed you guys fixed all the conflicts and it current state looks good.

Could you please confirm that's the case? If so we should merge it. Otherwise we should close it.

PR #106, opened 2023-05-06
Cluster Chart v0.0.1-alpha

itay-grudev:
Initial prototype of the `cluster` chart. 

The objective of the chart is to provide a really easy to use interface for cluster recovery and a convention over configuration approach.

It is the better approach in my opinion to simply a wrapper with respect to CNPG, as technically a CNPG cluster is just one resource that needs to be configured correctly.

Based on your feedback in #88 I designed the chart to operate in 3 `mode`s:
1. `standalone` - That would be your initial CNPG cluster
2. `recovery` - That's a cluster with a recovery bootstrap section
3. `replica` - For an off-site replica cluster (TODO)

There are three methods for recovery that you can choose from via `recovery.method`:
1. `backup`
2. `object_store`
3. `pg_basebackup` (TODO)

The chart by default separates the configuration between backup `barmanObjectStore` and recovery `barmanObjectStore` so that no conflicts of configuration arise.

I've implemented an `initial-backup` that is implemented as a `post-install` hook so that the cluster can provide PITR from its initialization (as opposed to after its first base or scheduled backup).

The chart includes an interface for specifying multiple `ScheduledBackups` and a PgBouncer.

I've implemented some base configuration for future out-of-the-box support of `PostGIS` via the `type` setting.

image

Closes: #178


Ornias1993:
Cool to see more people/projects working at helm-ifying the cluster portions.

We've just deployed a similair approach to this, baring any backup and recovery code for now for our whole catalog (truecharts) and the general design of putting this into helm works pretty well.

jouve:
FYI, I started something similar here https://github.com/jouve/charts/tree/main/charts/postgresql :)

itay-grudev:
@Ornias1993 The recovery code is the most important thing. You'd need a chart that not only works to setup your cluster, but helps you in recovery procedure when something goes wrong. The something went wrong part in my opinion is arguably the most important thing.

Ornias1993:
> @Ornias1993 The recovery code is the most important thing. You'd need a chart that not only works to setup your cluster, but helps you in recovery procedure when something goes wrong. The something went wrong part in my opinion is arguably the most important thing so your RTO stays low.

agreed.

phisco:
Hi, just wanted to chime in to say we value a lot your contributions, we evaluated many times in the past whether we wanted to create a chart for Clusters too, but we felt that would have been hard to to have something more than a 1:1 mapping with the Cluster resource, and we were not comfortable defining and maintaining any opinionated Cluster chart. So, it's awesome to see so many options coming out of the community and we are keeping this PR open for the moment just to see how this proposal evolves taking into account all the other options available, before officially committing and supporting it.
If you all want to have your charts built and hosted on your repos, we can help you set them up if you need, and then we could mention them in our docs as available options for Cluster charts and see if we can come up with a common denominator to officially support and add here.

charts/cluster/values.yaml
```diff
@@ -0,0 +1,211 @@
...
+  # Can be used to create databases, users, extensions, etc.
+  initSQL: []
+
+  additionalLabels: {}
+  annotations: {}
+
+###
+# Database cluster backup configuration
+backups:
+  enabled: true
```
phisco, line 162: enabling backups by default means that by default it will not work with the following error:
```
Error: execution error at (cluster/templates/cluster.yaml:75:5): You need to specify S3 region if endpointURL is not specified.
```
I would probably say this should be false and we should warn in the ` NOTES.txt` that it's highly discouraged running without backups, if `backups.enabled` is `false`

charts/cluster/templates/_bootstrap.yaml
```diff
@@ -0,0 +1,39 @@
...
+    {{- end }}
+
+externalClusters:
+  - name: objectStoreRecoveryCluster
+    barmanObjectStore:
+      serverName: {{ .Values.recovery.serverName }}
+      {{- $d := dict "chartFullname" (include "cluster.fullname" .) "scope" .Values.recovery "secretSuffix" "-recovery" -}}
+      {{- include "cluster.barmanObjectStoreConfig" $d | nindent 4 }}
+{{-  else }}
+  {{ fail "Invalid cluster mode!" }}
```
phisco, line 37: I see you are using fails, what about defining a [schema](https://helm.sh/docs/topics/charts/#schema-files) instead? you could define enums for these fields and avoid having to deal with wrong fields, unfortunately the tool we were using to automatically generate schemas was archived, but I forked it [here](https://github.com/phisco/helm-schema-gen) to make it run on ARM based macbooks too at least, but it doesn't support enums out of the box, so it's mainly going to be useful as a starting point, if you want below you can find the output of `helm schema-gen charts/cluster/values.yaml` with the current values.yaml:
```json
{
    "$schema": "http://json-schema.org/schema#",
    "type": "object",
    "properties": {
        "backups": {
            "type": "object",
            "properties": {
                "azure": {
                    "type": "object",
                    "properties": {
                        "connectionString": {
                            "type": "string"
                        },
                        "containerName": {
                            "type": "string"
                        },
                        "inheritFromAzureAD": {
                            "type": "boolean"
                        },
                        "path": {
                            "type": "string"
                        },
                        "serviceName": {
                            "type": "string"
                        },
                        "storageAccount": {
                            "type": "string"
                        },
                        "storageKey": {
                            "type": "string"
                        },
                        "storageSasToken": {
                            "type": "string"
                        }
                    }
                },
                "destinationPath": {
                    "type": "string"
                },
                "enabled": {
                    "type": "boolean"
                },
                "endpointURL": {
                    "type": "string"
                },
                "google": {
                    "type": "object",
                    "properties": {
                        "applicationCredentials": {
                            "type": "string"
                        },
                        "bucket": {
                            "type": "string"
                        },
                        "gkeEnvironment": {
                            "type": "boolean"
                        },
                        "path": {
                            "type": "string"
                        }
                    }
                },
                "provider": {
                    "type": "string"
                },
                "retentionPolicy": {
                    "type": "string"
                },
                "s3": {
                    "type": "object",
                    "properties": {
                        "accessKey": {
                            "type": "string"
                        },
                        "bucket": {
                            "type": "string"
                        },
                        "path": {
                            "type": "string"
                        },
                        "region": {
                            "type": "string"
                        },
                        "secretKey": {
                            "type": "string"
                        }
                    }
                },
                "scheduledBackups": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "backupOwnerReference": {
                                "type": "string"
                            },
                            "name": {
                                "type": "string"
                            },
                            "schedule": {
                                "type": "string"
                            }
                        }
                    }
                }
            }
        },
        "cluster": {
            "type": "object",
            "properties": {
                "additionalLabels": {
                    "type": "object"
                },
                "affinity": {
                    "type": "object",
                    "properties": {
                        "topologyKey": {
                            "type": "string"
                        }
                    }
                },
                "annotations": {
                    "type": "object"
                },
                "certificates": {
                    "type": "null"
                },
                "enableSuperuserAccess": {
                    "type": "boolean"
                },
                "imageName": {
                    "type": "string"
                },
                "imagePullPolicy": {
                    "type": "string"
                },
                "imagePullSecrets": {
                    "type": "array"
                },
                "initSQL": {
                    "type": "array"
                },
                "instances": {
                    "type": "integer"
                },
                "logLevel": {
                    "type": "string"
                },
                "monitoring": {
                    "type": "object",
                    "properties": {
                        "customQueries": {
                            "type": "array"
                        },
                        "enablePodMonitor": {
                            "type": "boolean"
                        },
                        "grafanaDashboard": {
                            "type": "object",
                            "properties": {
                                "enable": {
                                    "type": "boolean"
                                }
                            }
                        }
                    }
                },
                "postgresql": {
                    "type": "null"
                },
                "primaryUpdateMethod": {
                    "type": "string"
                },
                "primaryUpdateStrategy": {
                    "type": "string"
                },
                "priorityClassName": {
                    "type": "string"
                },
                "resources": {
                    "type": "null"
                },
                "storage": {
                    "type": "object",
                    "properties": {
                        "size": {
                            "type": "string"
                        },
                        "storageClass": {
                            "type": "string"
                        }
                    }
                },
                "superuserSecret": {
                    "type": "string"
                }
            }
        },
        "fullnameOverride": {
            "type": "string"
        },
        "mode": {
            "type": "string"
        },
        "nameOverride": {
            "type": "string"
        },
        "pooler": {
            "type": "object",
            "properties": {
                "enabled": {
                    "type": "boolean"
                },
                "instances": {
                    "type": "integer"
                },
                "parameters": {
                    "type": "object",
                    "properties": {
                        "default_pool_size": {
                            "type": "string"
                        },
                        "max_client_conn": {
                            "type": "string"
                        }
                    }
                },
                "poolMode": {
                    "type": "string"
                }
            }
        },
        "recovery": {
            "type": "object",
            "properties": {
                "azure": {
                    "type": "object",
                    "properties": {
                        "connectionString": {
                            "type": "string"
                        },
                        "containerName": {
                            "type": "string"
                        },
                        "inheritFromAzureAD": {
                            "type": "boolean"
                        },
                        "path": {
                            "type": "string"
                        },
                        "serviceName": {
                            "type": "string"
                        },
                        "storageAccount": {
                            "type": "string"
                        },
                        "storageKey": {
                            "type": "string"
                        },
                        "storageSasToken": {
                            "type": "string"
                        }
                    }
                },
                "backupName": {
                    "type": "string"
                },
                "clusterName": {
                    "type": "string"
                },
                "destinationPath": {
                    "type": "string"
                },
                "endpointURL": {
                    "type": "string"
                },
                "google": {
                    "type": "object",
                    "properties": {
                        "applicationCredentials": {
                            "type": "string"
                        },
                        "bucket": {
                            "type": "string"
                        },
                        "gkeEnvironment": {
                            "type": "boolean"
                        },
                        "path": {
                            "type": "string"
                        }
                    }
                },
                "method": {
                    "type": "string"
                },
                "pitrTarget": {
                    "type": "object",
                    "properties": {
                        "time": {
                            "type": "string"
                        }
                    }
                },
                "provider": {
                    "type": "string"
                },
                "s3": {
                    "type": "object",
                    "properties": {
                        "accessKey": {
                            "type": "string"
                        },
                        "bucket": {
                            "type": "string"
                        },
                        "path": {
                            "type": "string"
                        },
                        "region": {
                            "type": "string"
                        },
                        "secretKey": {
                            "type": "string"
                        }
                    }
                }
            }
        },
        "type": {
            "type": "string"
        }
    }
}
```

charts/cluster/templates/NOTES.txt
```diff
@@ -0,0 +1,13 @@
+The {{ include "cluster.fullname" . }} has been installed successfully.
+
```
phisco, line 2: maybe a summary of a few main configurations of the cluster could be helpful, nr of instances, image being used, backup configured or not (hinting at the fact that it should in production), schedules for backups and the backup destination, etc??

charts/cluster/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Things that are built but not tested. Remove corresponding lines after testing.
+
+* Azure Cloud Storage Backups
+* Google Cloud Storage Backups
+* Azure Cloud Storage Recovery
+* Google Cloud Storage Recovery
+
+## TODO
+* IAM Role for S3 Service Account
```
phisco, line 11: Do you have any plan yet about how to do this? I think it would be out of scope to actually set it up, but maybe we could hint at how to do that and provide the required configuration in the notes

charts/cluster/values.yaml
itay-grudev: I'd like to always have sane defaults for a safe, high-availability setup in the chart. I'd rather do it the other way around - a more descriptive error, when all fields are missing, that says:

```txt
A backup storage method hasn't been configured. You can disable backups if not needed using the `backups.enabled` flag.
```

itay-grudev: Would that be sufficient?

charts/cluster/templates/NOTES.txt
itay-grudev: That's a really good point. I'm on it.

charts/cluster/templates/_bootstrap.yaml
itay-grudev: I am not sure about he schema proposal. There are a lot of different validation configurations. I'd rather handle them in an:

```
if option1
else if option2
else
  fail
```

approach.

The problem is with nested checks configurations, for example certain fields need to be set only if backups are enabled. I think this way in general we can produce more helpful messages to guide the user.

Also it will be easier to maintain. Could we stick with this solution, at least until the chart has reached a stable version and we know we won't change configuration structure?

charts/cluster/values.yaml
phisco: It depends on the goal of the chart, if it's just for production clusters then it's ok to fail at first and make the message clearer and I would make it clearer, even through the name of the chart, if it should instead be a more handheld experience to spinning up a Cluster at first locally and then up to production, I'd allow an unsafe config by default and add warnings

charts/cluster/templates/_bootstrap.yaml
phisco: Yes, you'd still need to specify those conditional checks via the chart, I was mostly suggesting it for the enums and plain required fields, to avoid having the else branches dealing with invalid values. But given that it would be mostly a manual thing I agree we can postpone that to when the values will be more stable 👍

charts/cluster/templates/NOTES.txt
itay-grudev: image

charts/cluster/values.yaml
itay-grudev: I've went with your suggestion and have disabled backups by default. It's also backed by a warning message in the NOTES.txt. See the image below.

charts/cluster/README.md
itay-grudev: I've improved the README to include information on what's tested and what isn't along with a big warning message.

I've also included a note on the purpose of the chart based on yours and Jonathan's feedback in the CNPG meeting.

And finally, I've asked and @Otterpohl kindly agreed to look into the IAM Role functionality as I currently don't have the capability to do so.

charts/cluster/templates/NOTES.txt
phisco: This looks awesome! 🎉

charts/cluster/templates/cluster.yaml
```diff
@@ -0,0 +1,65 @@
...
+  {{- with .Values.cluster.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  labels:
+  {{- include "cluster.labels" . | nindent 4 }}
+  {{- with .Values.cluster.additionalLabels }}
+    {{ toYaml . | nindent 4 }}
+  {{- end }}
+spec:
```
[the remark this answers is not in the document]
itay-grudev: A separate PR for that is better. This one is already pretty large.

charts/cluster/templates/cluster.yaml
```diff
@@ -0,0 +1,65 @@
...
+  instances: {{ .Values.cluster.instances }}
```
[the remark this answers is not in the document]
itay-grudev: A separate PR for that is better as well. This one is already pretty large.

charts/cluster/templates/initial-backup.yaml
```diff
@@ -0,0 +1,17 @@
+{{ if .Values.backups.enabled }}
+apiVersion: postgresql.cnpg.io/v1
+kind: Backup
+metadata:
+  name: {{ include "cluster.fullname" . }}-initial-backup-{{now | unixEpoch}}
+  annotations:
+    "helm.sh/hook": post-install
```
Ornias1993, line 7: Enabling backups after initial install won't create an initial backup.

Best solution, though nog argoCD compatible, would be to do a lookup and skip creating a backup if one with said name already exists combined with helm.sh/resource-policy: keep to prevent deletion using helm.

armru:
@phisco @mnencia @gbartolini @sxd  what do you think is missing to merge it?

itay-grudev:
I rebased all the changes to sync them up with `main`.

Ornias1993:
We expect to have our variant as part of our common library chart basically viable within a month or so and backup Q1 2024 :)

charts/cloudnative-pg/Chart.yaml
```diff
@@ -15,7 +15,7 @@
 #
 apiVersion: v2
 name: cloudnative-pg
-description: CloudNativePG Helm Chart
+description: CloudNativePG Operator Helm Chart
```
Ornias1993, line 18:
```suggestion
description: CloudNativePG ClusterHelm Chart
```

itay-grudev: Uhmm.. That's the operator chart: `charts/cloudnative-pg/Chart.yaml`, as opposed to ``charts/cluster/Chart.yaml`. The suggestion doesn't make sense.

revolunet:
Thanks ! 
About the initial-backup, is it necessary to use a post-install hook?
Looks like [ScheduledBackup](https://cloudnative-pg.io/documentation/1.21/cloudnative-pg.v1/#postgresql-cnpg-io-v1-ScheduledBackupSpec) has a `.spec.immediate` parameter

charts/cloudnative-pg/Chart.yaml
Ornias1993: Ohh shoot, misread sorry!

Ornias1993:
> Thanks ! About the initial-backup, is it necessary to use a post-install hook? Looks like [ScheduledBackup](https://cloudnative-pg.io/documentation/1.21/cloudnative-pg.v1/#postgresql-cnpg-io-v1-ScheduledBackupSpec) has a `.spec.immediate` parameter

No, post install objects wont get deleted on deletion of the helm-chart, putting each and every helm chart using this, into a potential limbo when users try to remove the helm chart and their namespace (it would lead to the namespace being put into deletion limbo).

Known issue, reported many times.
Also fix has been reported (allow deletion on chart deletion with a flag).

But Helm developers are terribly slow and basically AWOL 90% of the time.

itay-grudev:
@Ornias1993 That makes sense. I'll make the change.

sxd (approved):

phisco (approved):

Ornias1993:
Last month we've implemented a variant of this, fully integrated in the TrueCharts Common-Chart.
Since then we've covered it withsignificant unittest coverage as well and added GUI components to have the backup functionality also available on TrueNAS SCALE Apps.

Since then it has been running on our thousands of users without issue.

I would say, however, that for charts at this scope, unittests are really something you should consider before merging even.

revolunet:
A JSON schema would be a nice addition to this PR

itay-grudev:
> A JSON schema would be a nice addition to this PR

I've been think about the suggestion.

An accurate JSON schema that reflects changes in the operator might be really difficult to maintain. A simple one that only reflects the overall structure of the chart might not add much in terms of safety.

In any way, it is for a point for future PR.

itay-grudev:
Closed in favor of #179

PR #118, opened 2023-06-05
feat(cloudnative-pg): clusterrole aggregated to user facing roles

jouve:
create rbac to view/edit cnpg crds and aggregate to [user facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles)

view/edit/admin clusterroles are bound to a user & namespace, to manage said namespace, and now allow to create a cluster/backup/etc  with this change.


phisco:
Can you remove the bump to the version? Otherwise looks good to me, thanks for switching to false by default 🙏

phisco (approved):

itay-grudev:
Is this still relevant or should we close it?

@jouve

jouve:
forgot about this one, it was implemented by #39

PR #126, opened 2023-07-04
Proposal to add structured field monitoringQueriesConfigMap.queriesData

georgebarbarosie:
Uses same syntax as `queries`, but it's plain Yaml instead of yaml-as-string. The values in both fields are merged, with `queriesData` overriding values from `queries`.

We pass queries through `fromYaml`, we then merge it with `queriesData` and use the result as the final value in the ConfigMap.

The advantage of this is that it allows people to override specific queries in the default queries value and it allows composability for those that use e.g. kustomizations to compile the values of the chart (gitops, fluxcd, etc).


phisco:
@georgebarbarosie can you check the error with the DCO? https://github.com/cloudnative-pg/charts/runs/14778552401

georgebarbarosie:
> @georgebarbarosie can you check the error with the DCO? https://github.com/cloudnative-pg/charts/runs/14778552401

Done. Apologies, I wasn't aware of the requirement.

PR #141, opened 2023-09-08
Move CRDs to helm's crds folder

edeustua:
Helm v3 [recommends putting all CRDs in its `crds` folder](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) on the root of the chart. This ensures that all CRDs are installed before anything else. Right now, installing cnpg as a subchart dependency fails when manifests containing cnpg's CRDs are present. To avoid this issue, we are moving the all crds from their present location at `templates/crds` to `crds`.


phisco:
We update crds in place, which AFAIK is something helm won't allow us to do. This would be a breaking change for the rest of the people installing the operator and Clusters separately as they would have to find a workaround to upgrade the CRDs.

edeustua:
Also, what about having them both? That way you get them installed at the right time when solving the dependencies and one can also update in place?

edeustua:
This PR doesn't make sense for now.

PR #148, opened 2023-10-10
feat(cloudnative-pg): Configuration of PodMonitor relabeling rules

npdgm:
Hi,
Adding a new chart value to configure Prometheus relabeling rules through the PodMonitor.
This facility is often used for renaming or dropping labels.
Cheers


phisco:
Hi @npdgm! Thanks a for your contribution. Just to make sure, this is going to affect only the operator's metrics, not the Clusters' ones, that's what you wanted right?

npdgm:
Yes I'm aware of that. Probably you saw in the issue I quickly closed about Cluster metric labels :)
And indeed the next thing I'll need is to look into operator managed PodMonitors and see how relabelings could fit there.

itay-grudev (approved):

sxd (approved):

PR #166, opened 2023-11-18
Cnpg operator renamechart

wibed:
cloudnative-pg is too long for a name including a dash in its most basic form.
making any extension cumbersome and hard to look at.

cnpg for short seems fine to me. 

it allows for charts like cnpg-db and cnpg-operator to live side by side within the same "namespace" without polluting the naming scheme with confusing additional inecessary dashes.
as to be found within:
- cloudnative-pg-operator
- cloudnative-pg-db


bjw-s:
Not sure what it is you are trying to do here, but either it is a mistake, or you are missing a useful PR description...

wibed:
> Not sure what it is you are trying to do here, but either it is a mistake, or you are missing a useful PR description...

please it felt obvious to me. thanks for the clarification

onedr0p:
Personally I think the naming is fine as it is.

However, If you really wanted to make this change this isn't how you would go about it. 
You would need to [deprecate the existing chart](https://github.com/helm/charts/blob/master/PROCESSES.md), not rename it. Once that happens and it's released then you can create a PR to add a new chart with the updated name.

phisco:
+1 to what the other said above.

As a general rule of thumb, I would suggest first opening an issue to discuss things like this in the future.

wibed:
> +1 to what the other said above.
> 
> As a general rule of thumb, I would suggest first opening an issue to discuss things like this in the future.

Definitively! if there are any suggestion please feel free to suggest them.

wibed:
what is the state of this pr?

itay-grudev:
If anyone is seriously considering a rename, go with simply `operator` so that the important charts in the repository would be:

* `cnpg/operator`  
* `cnpg/cluster`

That being said there is really no practical need for a rename and that would only cause issues for existing users.

wibed:
> If anyone is seriously considering a rename, go with simply `operator` so that the important charts in the repository would be:
> 
> * `cnpg/operator`
> * `cnpg/cluster`
> 
> **That being said there is really no practical need for a rename and that would only cause issues for existing users.**

i disagree. keeping cnpg in the name distinguishes instances between each other.
it is very inconvenient to go search for identifying bits of information when the name is only "operator"

this pr is ground enough to discuss. everyone can post his opinion

onedr0p:
> this pr is ground enough to discuss. everyone can post his opinion

I disagree, major changes like should be discussed in a Github issue and agreed upon by the cnpg maintainers before work is started on them.

PR #174, opened 2024-01-17
fix: set version well-known label with specified value

### LINKED ISSUE #173: version well-know label is hardcoded to appVersion in Chart.yaml
The intention seems to be that it should match what is set by `.Values.image.tag` else default to `.Chart.appVersion`.

This issue causes changes to the label even when `.Values.image.tag` is set ti an older value (as of 17th Jan 2024):

```diff
-    app.kubernetes.io/version: 1.20.2
-    helm.sh/chart: cloudnative-pg-0.18.2
+    app.kubernetes.io/version: 1.22.0
+    helm.sh/chart: cloudnative-pg-0.20.0
```

nareshov:
Fixes #173 

before:
```
❯ helm template test charts/cloudnative-pg | grep app.kubernetes.io/version                       
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"

❯ helm template test charts/cloudnative-pg --set image.tag=1.20.2 | grep app.kubernetes.io/version
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
```

after:
```
❯ helm template test charts/cloudnative-pg | grep app.kubernetes.io/version
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"
    app.kubernetes.io/version: "1.22.0"

❯ helm template test charts/cloudnative-pg --set image.tag=1.20.2 | grep app.kubernetes.io/version
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
    app.kubernetes.io/version: "1.20.2"
```


charts/cloudnative-pg/templates/_helpers.tpl
```diff
@@ -36,9 +36,7 @@ Common labels
 {{- define "cloudnative-pg.labels" -}}
 helm.sh/chart: {{ include "cloudnative-pg.chart" . }}
 {{ include "cloudnative-pg.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
```
nareshov, line 39: dropped this if conditional seeing as it appears to not be necessary anymore - unless I'm missing something non-obvious.

nareshov:
@itay-grudev @phisco - poke poke 🙏🏾

phisco:
That's a default from Helm, when you create a new chart using `helm create my_chart`, you get a set of helpers, one of them is for common labels and the version is set to the AppVersion.

Looking at other widely used charts it looks like no one is changing that. Which I think makes, as the image tag has nothing to do with the version of the app, the tag could be whatever you like (in our case it has to be a semantic version, because we use that, but for most other projects it doesn't) and if that's actually an older or more recent version of the application the chart was meant for, even if it might work, that's not the version the chart was meant to work with.

itay-grudev:
Actually @phisco I think @nareshov might have a point. According to the docs:

> app.kubernetes.io/version - **The current version of the application** (e.g., a [SemVer 1.0](https://semver.org/spec/v1.0.0.html), revision hash, etc.)

The chart here supports overriding the operator version by specifying a different `image.tag`, and it genuinely makes sense to override the labels.

As to the argument that this is a Helm default - it is - one that makes sense as a default. Not necessarily the best implementation. I get gunienly upset by people doing stuff a certain way only because everyone else is. This way we might end up in a local maximum within the field of success.

phisco:
Exactly:

> app.kubernetes.io/version - The current version of the application (e.g., a SemVer 1.0, revision hash, etc.)

What you mean by application is the key point here. `AppVersion` is the application version, so it looks fitting here. That default assumes the helm chart IS the application, which I agree with. And, as I was saying, the image tag has little to do with the actual version being deployed, you should not use the chart with a version of the operator different from the one it was meant for.

itay-grudev:
This is the chart version:
https://github.com/cloudnative-pg/charts/blob/42ab86f7be5d65df87ae03cce255e8ff6a1905a7/charts/cloudnative-pg/Chart.yaml#L21

This is the application version running inside the chart:
https://github.com/cloudnative-pg/charts/blob/42ab86f7be5d65df87ae03cce255e8ff6a1905a7/charts/cloudnative-pg/Chart.yaml#L26

And importantly, the helm chart is not application. The CNPG operator it is provisioning is. And just to prove that there is a precedent for this, here is an excerpt from the **Grafana Helm chart**:

https://github.com/grafana/helm-charts/blob/674b94682f1a2ebf8a72ca9a698f181b52c05592/charts/grafana/templates/_helpers.tpl#L64

itay-grudev:
@nareshov I think @phisco's argument makes a lot of sense. In fact I think it makes much more sense to either:

* Remove `image.tag` overrides, although that could lead to problems on its own with users who need to host their own docker image.

* Indicate explicitly that the version is overridden and outside of normal operating parameters. Consider this for example:
```yaml
app.kubernetes.io/version: {{ ternary .Chart.AppVersion (printf "0.0.0-%s" .Values.image.tag) (empty .Values.image.tag) | quote }}
```

This would produce the following:

```yaml
app.kubernetes.io/version: "1.22.2" # when no image.tag is set
app.kubernetes.io/version: "0.0.0-yourtag" # when an override is specified
```

Here is a **[repl](https://repeatit.io/#/share/eyJ0ZW1wbGF0ZSI6Int7IFwiXCIgfCBkZWZhdWx0IFwiYmxhXCIgfX1cbnt7IFwiYXNkXCIgfCBkZWZhdWx0IFwiYmxhXCIgfX1cblxue3sgZGVmYXVsdCBcImJsYVwiIFwiXCIgfX1cbnt7IGRlZmF1bHQgXCJibGFcIiBcImFzZFwiIH19IiwiaW5wdXQiOiIiLCJjb25maWciOnsidGVtcGxhdGUiOiJ0ZXh0IiwiZnVsbFNjcmVlbkhUTUwiOmZhbHNlLCJmdW5jdGlvbnMiOlsic3ByaWciXSwib3B0aW9ucyI6WyJsaXZlIl0sImlucHV0VHlwZSI6InlhbWwifX0=)**.

I also now think this MR should remain closed.

Issue #178, opened 2024-02-01
Provide a Cluster chart


sxd:
We have an operator chart, but we lack a chart that initializes a database cluster.

The objective of such a chart should not be to provide a straight through interface to a `Cluster.postgresql.cnpg.io` object. Advanced use should be done by interacting with CNPG's CRDs directly. Instead to the chart should provide a reliable, easy-to-use method for setting up a CNPG cluster and related resources with safe defaults. 

An good cluster chart should cover multiple aspects, such as:
* Provision a cluster
* Provision a Pooler
* Backup configuration
* Scheduled Backups
* Disaster recovery
* Alerting and Monitoring
* Safe and reliable defaults and warnings whenever user settings result in unsafe configuration
* Tests

jwklijnsma:
This is nice one we made are own 8 months ago

PR #179, opened 2024-02-02
Cluster Chart v0.0.1-alpha

sxd:
Initial prototype of the `cluster` chart. 

The objective of the chart is to provide a really easy to use interface for cluster recovery and a convention over configuration approach.

It is the better approach in my opinion to simply a wrapper with respect to CNPG, as technically a CNPG cluster is just one resource that needs to be configured correctly.

Based on your feedback in #88 I designed the chart to operate in 3 `mode`s:
1. `standalone` - That would be your initial CNPG cluster
2. `recovery` - That's a cluster with a recovery bootstrap section
3. `replica` - For an off-site replica cluster (TODO)

There are three methods for recovery that you can choose from via `recovery.method`:
1. `backup`
2. `object_store`
3. `pg_basebackup` (TODO)

The chart by default separates the configuration between backup `barmanObjectStore` and recovery `barmanObjectStore` so that no conflicts of configuration arise.

I've implemented an `initial-backup` that is implemented as a `post-install` hook so that the cluster can provide PITR from its initialization (as opposed to after its first base or scheduled backup).

The chart includes an interface for specifying multiple `ScheduledBackups` and a PgBouncer.

I've implemented some base configuration for future out-of-the-box support of `PostGIS` via the `type` setting.

image

Closes: #178


.github/actions/setup-kind/action.yml
```diff
@@ -0,0 +1,24 @@
...
+description: Sets up a kind cluster and installs Helm and kubectl
+outputs:
+  helm-path:
+    description: The path to the Helm binary
+    value: ${{ steps.helm.outputs.helm-path }}
+  kubectl-path:
+    description: The path to the kubectl binary
+    value: ${{ steps.kubectl.outputs.kubectl-path }}
+runs:
+  using: "composite"
```
phisco, line 11:
```suggestion
  using: composite
```

phisco (approved):

itay-grudev (approved):

PR #191, opened 2024-02-22
Added PGP Key for Helm provenance

### LINKED ISSUE #163: [Security]: Provide Helm Provenance to mitigate supply chain attacks
It would be nice if automated Helm-Provenance is provided with the Chart.
This ensures users can verify the origin of the Helm-Charts and prevents some supply chain attacks.
More info: https://helm.sh/docs/topics/provenance/

Even more so when bundled with signing the images as well.

--

As a downstream, the current unsecure distribution prevents us from directly referencing these Helm charts, as it doesn't qualify our security policy. Which requires correct signing of images and provenance on helm charts.

itay-grudev:
Before merging:

Copy the private key block to a GitHub Repository Secret called: `PGP_PRIVATE_KEY` , then delete it.
Copy the key encryption password to a GitHub Secret called: `PGP_KEY_PASSPHRASE`.

Closes #163


phisco (approved):

.github/workflows/release-publish.yml
```diff
@@ -27,6 +27,11 @@ jobs:
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
+      - name: Import PGP Private Key
+        run: |
+          echo "${{ secrets.PGP_PRIVATE_KEY }}" | gpg --dearmor --output keyring.gpg
+          echo "${{ secrets.PGP_KEY_PASSPHRASE }}" > passphrase-file.txt
```
phisco, line 33: maybe passing it through the env would be better, see https://github.com/keptn/lifecycle-toolkit-charts/blob/3de78103af8f8f3690c3a0106508c036a50ac56a/.github/workflows/release.yml#L40C1-L42C1

phisco: I now noticed in the example above they are also creating it as a file and just passing the filepath... nevermind 😅

phisco (approved):
Thanks @itay-grudev!

Ornias1993:
Thats some quick work @itay-grudev and @phisco Looks solid!

Issue #195, opened 2024-02-25
Add support for additional labels in PodMonitor


cvrajeesh:
Currently, there is no way to add extra labels for `PodMonitor`. This is useful when they're discovered using specific labels e.g. `kube-prometheus-stack` expects a label `release: kube-prometheus-stack` on the PodMonitor.

Suggestions

1. New key called `podMonitorAdditionalLabels`

```yaml
monitoring:
  podMonitorEnabled: true
  podMonitorAdditionalLabels:
    release: kube-prometheus-stack
```

2. Dreprecate `podMonitorEnabled` instead and create a new key called `podMonitor`

```yaml
monitoring:
  podMonitor:
    enabled: true
    additionalLabels:
       release: kube-prometheus-stack
```

This needs to be done for both `cloudnative-pg` and `cluster` charts

cvrajeesh:
Can't do this on the `cluster` charts, unless this is fixed https://github.com/cloudnative-pg/cloudnative-pg/discussions/1617

itay-grudev:
@cvrajeesh I've bypassed this issue by setting the following in my `kube-prometheus-stack` installation so that it no longer looks for a release label:

```yaml
prometheus:
  prometheusSpec:
    podMonitorSelectorNilUsesHelmValues: false
    ruleSelectorNilUsesHelmValues: false
    serviceMonitorSelectorNilUsesHelmValues: false
    probeSelectorNilUsesHelmValues: false
```

That being said, if this is still an issue in the operator we should raise a ticket from the https://github.com/cloudnative-pg/cloudnative-pg/discussions/1617 discussion that the `Cluster` labels are not replicated to the `PodMonitor`.

itay-grudev:
P.S. We already renamed the values for the monitoring section:
```yaml
cluster:
  monitoring:
    enabled: false
    podMonitor:
      enabled: true
```

PR #196, opened 2024-02-25
feat(cloudnative-pg): Support for additional labels in PodMonitor

cvrajeesh:
Add support for additional labels to `PodMonitor`

```yaml
monitoring:
  podMonitor:
    enabled: true
    additionalLabels:
       release: kube-prometheus-stack
```

Closes: https://github.com/cloudnative-pg/charts/issues/195


itay-grudev:
We are very strict about backwards compatibility on anything that we consider stable.

As such however prettier, we unfortunately cannot accept the breaking change of renaming `podMonitorEnabled` to `podMonitor.enabled`.

As such, the next best thing would be to call the labels: `podMonitorAdditionalLabels`.

itay-grudev (approved):

PR #199, opened 2024-02-29
feat: allow using existing secret for backup and restore

### LINKED ISSUE #197: Add support for existing backup credentials

DASPRiD:
This PR allows to specify S3 credentials through an existing secret instead of having to commit the secrets with custom helm chart values.

Closes #197

### DIFF
--- a/charts/cluster/templates/_barman_object_store.tpl
+++ b/charts/cluster/templates/_barman_object_store.tpl
@@ -15,33 +15,35 @@
   {{- if empty .scope.destinationPath }}
   destinationPath: "s3://{{ required "You need to specify S3 bucket if destinationPath is not specified." .scope.s3.bucket }}{{ .scope.s3.path }}"
   {{- end }}
+  {{ $secretName := coalesce .scope.secret.name (printf "%s-backup-s3%s-creds" .chartFullname .secretSuffix) }}
   s3Credentials:
     accessKeyId:
-      name: {{ .chartFullname }}-backup-s3{{ .secretSuffix }}-creds
+      name: {{ $secretName }}
       key: ACCESS_KEY_ID
     secretAccessKey:
-      name: {{ .chartFullname }}-backup-s3{{ .secretSuffix }}-creds
+      name: {{ $secretName }}
       key: ACCESS_SECRET_KEY
 {{- else if eq .scope.provider "azure" }}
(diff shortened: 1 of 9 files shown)


itay-grudev:
Can you apply the same functionality to the google and azure secrets as well for consistency?

itay-grudev:
Also please make sure to `signoff` your commits and amend your existing commit(s) to have a signoff as well, for the DCO check.

phisco (commented):
You might want to have `secret.create` (default true) and `secret.name` (default empty) so that one can both specify a custom name for the secret. It's pretty common.
I wonder if it could be moved to an upper level so that we don't have to repeat it for all flavors.

DASPRiD:
> Can you apply the same functionality to the google and azure secrets as well for consistency?

I did look at that initially, but I'm not sure how exactly to handle that with Azure, since it has if statements depending on the auto-created secret.

charts/cluster/templates/_barman_object_store.tpl
```diff
@@ -15,44 +15,47 @@
...
-      name: {{ .chartFullname }}-backup-s3{{ .secretSuffix }}-creds
+      name: {{ $secretName }}
       key: ACCESS_SECRET_KEY
 {{- else if eq .scope.provider "azure" }}
   {{- if empty .scope.destinationPath }}
   destinationPath: "https://{{ required "You need to specify Azure storageAccount if destinationPath is not specified." .scope.azure.storageAccount }}.{{ .scope.azure.serviceName }}.core.windows.net/{{ .scope.azure.containerName }}{{ .scope.azure.path }}"
   {{- end }}
   azureCredentials:
+  {{ $secretName := coalesce .scope.secret.name (printf "%s-backup-azure%s-creds" .chartFullname .secretSuffix) }}
   {{- if .scope.azure.connectionString }}
```
DASPRiD, line 32: I need some guidance here. Since the credentials are not defined in the values anymore, how would you suggest to handle this?

DASPRiD:
Alright, I signed off the commit and implemented the suggested parent secret property for both recovery and backups. I'm still stumbled on the azure config, see my inline comment.

itay-grudev:
I reworked your proposal in #239.

Closing in favor of #239.

PR #205, opened 2024-03-08
Adds the possibility to write annotations to the Grafana dashboard

0hlov3:
If you use the Prometheus Operator Helm Chart to Deploy the Monitoring Stack, there is the Possibility to import your Dashboards to Grafana. The Grafana Sidecar can scrape Labels and Annotations from the Configmap. This adds the Possibility to create Annotations for the Grafana Dashboard, in my use case, i want to create different Dashboard folder, so things are sorted.

Example values.yaml:
```shell
monitoring:
  # -- Specifies whether the monitoring should be enabled. Requires Prometheus Operator CRDs.
  podMonitorEnabled: true
  grafanaDashboard:
    create: true
    sidecarLabel: "grafana_dashboard"
    # -- Label value that ConfigMaps should have to be loaded as dashboards.
    sidecarLabelValue: "1"
    # -- Annotations that ConfigMaps should have to get Configured.
    annotations:
      grafana_dashboard_folder: CNPG
``` 

Which will create a Folder in Grafana view.

I would love, if we could merge this, I am also happy to implement suggestions for improvement.


charts/cloudnative-pg/values.yaml
```diff
@@ -149,6 +149,8 @@ monitoring:
     sidecarLabel: "grafana_dashboard"
     # -- Label value that ConfigMaps should have to be loaded as dashboards.
     sidecarLabelValue: ""
+    # -- Annotations that ConfigMaps should have to get Configured.
```
phisco, line 152: What do you mean by "to get configured"?

0hlov3: Sorry, I did a Mistake in the Description, but i don't know how to write the correct description .

I want to say:
# -- Annotations that ConfigMaps can have to get configured in grafana, for example:
## grafana_dashboard_folder: CNPG will Create a Folder CNPG in Grafanas dashboard overview and import the the Dashboard into this folder.

.gitignore
```diff
@@ -1,3 +1,5 @@
+# Ignores any `development` Folder
+**/**/development/
```
itay-grudev, line 2: I assume this is something you used internally. Can you please remove it?

itay-grudev (changes requested):
At this point for consistency we might as well just have a `labels` field that defaults to:

```yaml
labels:
  grafana_dashboard: "1"
```

And remove `sidecarLabel` and `sidecarLabelValue`.

@phisco Do you approve?

.gitignore
0hlov3: Yes sorry, that was only for development purposes, so that i don't accidentally push something, i don't want to. Removed it from the history.

phisco:
@itay-grudev wouldn't that be a breaking change though?

itay-grudev:
> @itay-grudev wouldn't that be a breaking change though?

Yhea. Bad idea.

0hlov3:
> sidecarLabelValue

I can change sidecarLabelValue to default to 1, is there any thing to do from my side? 

Thanks! :)

itay-grudev (approved):

itay-grudev:
Closed in favor of #225

PR #210, opened 2024-03-13
Fix cluster resources value type

gppmilicia:
Hi,
I was trying to configure the cluster resources for the pods, and I got this error:

```
cluster:
    - cluster.resources: Invalid type. Expected: null, given: object
```

This error seems to be related only to the wrong mapping in `value.schema.json`, which is fixed in this PR.

I added an example that is the same one I was trying to apply in my cluster.


itay-grudev:
Closed in favor of #211

PR #211, opened 2024-03-13
Bug Fix: Schema shows resources as null

itay-grudev:
Closes #210

### DIFF
--- a/charts/cluster/values.schema.json
+++ b/charts/cluster/values.schema.json
@@ -173,14 +173,14 @@
                         }
                     }
                 },
-                "postgresql": {
-                    "type": "object"
+                "postgresGID": {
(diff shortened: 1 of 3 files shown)


charts/cluster/values.schema.json
```diff
@@ -173,14 +173,14 @@
...
+                "postgresGID": {
+                    "type": "integer"
                 },
                 "postgresUID": {
                     "type": "integer"
                 },
-                "postgresGID": {
-                    "type": "integer"
+                "postgresql": {
+                    "type": "object"
```
itay-grudev, line 183: These were modified manually and the generator just reorders them.

phisco (approved):

PR #216, opened 2024-03-14
feat: add annotation to grafana dashboard

ishioni:
Add grafana_folder annotation to the dashboard which allows it to be automagically placed in a folder


itay-grudev:
Closed in favor of #205.

Issue #198, opened 2024-02-28
Backups incompatibility with unencrypted minio


donch:
When i enable S3 backups with a minio configuration.

The barmanConfig looks like : 

`  barmanObjectStore:
    wal:
      compression: gzip
      encryption: AES256
    data:
      compression: gzip
      encryption: AES256
      jobs: 2`
      
This configuration doesn't work if you don't enable KMS/Encryption on your minio. It would be nice to configure these settings.
For exemple, we need to change the AES256 to "" for backup and wal archiving to work.

itay-grudev:
Am I correct in assuming that all you are asking for is that the settings are exposed in `.Values`?

itay-grudev:
@donch Would #221 do the trick?

PR #221, opened 2024-03-17
feat(cluster): Exposed backup.barmanObjectStore.wal and backup.barmanObjectStore.data

itay-grudev:
This PR exposed the `backup.barmanObjectStore.wal` and `backup.barmanObjectStore.data` properties so that the encryption can be disabled.

This is needed in the case where you are using MinIO without encryption enabled. The defaults for encryption are still: `AES256`.

In addition the compression and parallelism properties are exposed for fine-tuning the storage space and performance of your cluster.

Closes: #198


phisco (approved):
LGTM 🙏

PR #225, opened 2024-03-21
Imeplemented the Grafana dashboard as a dependency

itay-grudev:
Closes #205

The CI will fail until we finish the Grafana Dashboards repository. And after it's published we need to update the `Chart.lock` here.


charts/cloudnative-pg/values.yaml
```diff
@@ -145,10 +145,15 @@ monitoring:
...
+    # -- Label value that ConfigMaps should have to be loaded as dashboards.  DEPRECATED: Use labels instead.
     sidecarLabelValue: "1"
+    # -- Labels that ConfigMaps should have to get configured in Grafana.
+    labels:
+      grafana_dashboard: "1"
+    # -- Annotations that ConfigMaps can have to get configured in Grafana.
+    annotations: {}
 
 # Default monitoring queries
 monitoringQueriesConfigMap:
```
phisco, line 159: Is this going to be passed to the dependency because we set monitoring as alias for it? 🤔

itay-grudev: Precisely. Because there is no way to access the template from the parent chart. It needs to be generated by the subchart.

.github/actions/deploy-operator/action.yml
```diff
@@ -5,6 +5,12 @@ runs:
...
-        --create-namespace charts/cloudnative-pg --wait
+      run:
+        helm dependency update charts/cloudnative-pg
+
+        helm upgrade
+        --install 
+        --namespace cnpg-system
+        --create-namespace
+        --wait
+        cnpg charts/cloudnative-pg
```
phisco, line 16: I think this won't work otherwise, right? 🤔
```suggestion
        helm upgrade \
          --install \
          --namespace cnpg-system \
          --create-namespace \
          --wait \
          cnpg charts/cloudnative-pg
```

phisco: looks like it still works 🤔

phisco (approved):
Awesome, @itay-grudev 🙏

PR #226, opened 2024-03-22
feat(cluster): Make roles configurable

### LINKED ISSUE #201: Add support for declarative role management
Currently, the cluster does not allow assigning the "managed" section to manage user roles and credentials, as indicated in: 

https://cloudnative-pg.io/documentation/1.22/declarative_role_management/

gpothier:
This makes it possible to configure roles using the chart by adding an array in `cluster.roles` in `values.yaml`. The content of the array are taken as is and used as `.spec.managed.roles`.

Closes #201


itay-grudev (changes requested):
Can you also:
1. Pull with the rebase I just made.
2. Signoff on your changes (the [DCO](https://github.com/cloudnative-pg/charts/pull/226/checks) check)
3. Run `make schema docs`

gpothier:
Hi, thanks for taking a look. This week I'm on holiday so I'll be able to
do the pull and the make schema docs next week. What do you mean by signoff
on my changes?

Regarding putting the `with` before `managed`, I opted to put it inside in
case it is needed at some point to bring other things into `managed`. In
that case we would have to put an `if` with several conditions outside, and
several `ifs` inside in order to avoid the empty element. IIRC I did try
the chart without specifying roles to check if the empty `managed` was a
problem, and there was no problem.

Regards,
g

gpothier:
Hi, I ran `make schema docs`, but it looks like it is deleting the contents of `charts/cloudnative-pg/values.schema.json` and `charts/cluster/values.schema.json`, I suppose this is not right? I have not committed that.

gpothier:
Hi, I just noticed the link to the DCO check, so I did that, but now I notice there are a lot of conflicts, I don't understand why (very very basic git user here). Would it be easier if I close this PR and create a new one?

itay-grudev:
Yhea - that's not a bad idea. Something happened with that last force-push but I can't tell what.

PR #232, opened 2024-03-26
Added the ability to exclude specific PrometheusRules

itay-grudev:
Adds the ability to disable specific PrometheusRules:

```yaml
cluster:
  monitoring:
    prometheusRule:
      excludeRules: # NEW
        - CNPGClusterHAWarning
        - CNPGClusterZoneSpreadWarning
```

In the process I also restructured how Prometheus rules are generated so each one now has its own file.

### DIFF
--- a/charts/cluster/prometheus_rules/cluster-ha-critical.yaml
+++ b/charts/cluster/prometheus_rules/cluster-ha-critical.yaml
@@ -1,0 +1,23 @@
+{{- if not (has "CNPGClusterHACritical" .excludeRules) -}}
+alert: CNPGClusterHACritical
+annotations:
+  summary: CNPG Cluster has no standby replicas!
+  description: |-
+    CloudNativePG Cluster "{{ .labels.job }}" has no ready standby replicas. Your cluster at a severe
+    risk of data loss and downtime if the primary instance fails.
+
+    The primary instance is still online and able to serve queries, although connections to the `-ro` endpoint
+    will fail. The `-r` endpoint os operating at reduced capacity and all traffic is being served by the main.
+
+    This can happen during a normal fail-over or automated minor version upgrades in a cluster with 2 or less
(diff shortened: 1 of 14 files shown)


charts/cluster/prometheus_rules/cluster-ha-critical.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if not (has "CNPGClusterHACritical" .excludeRules) -}}
+alert: CNPGClusterHACritical
```
phisco, line 2: Nit: you can define a variable with the name and use it both as name and in the condition.

itay-grudev: AAAAAAHhhh... to avoid repeating the name.. Got it.

itay-grudev: @phisco Done!

charts/cluster/values.schema.json
```diff
@@ -213,6 +213,9 @@
                             "properties": {
                                 "enabled": {
                                     "type": "boolean"
+                                },
+                                "excludeRules": {
+                                    "type": "array"
```
phisco, line 218: We could define the array to be of strings and define all the possible values, but maybe it's a bit overkill

itay-grudev: The moment we start doing that we have to throw away the schema generator, because it will overwrite the customisations.

itay-grudev: Not that I love it, but it does offer some convenience.

phisco: Yeah, I know, it would be nice to have annotations similar to kubebuilder, but I'm not sure I'll have any time to work on it in the foreseeable future 😅

phisco (approved):

PR #239, opened 2024-03-28
feat(cluster): allow using existing secret for backup and restore

### LINKED ISSUE #197: Add support for existing backup credentials

itay-grudev:
Closes #197


itay-grudev:
@phisco I've made your review slightly more complicated as this is the base for my work on adding support for replica clusters and I've fixed and restructured the credential secrets.

gpothier:
Hi, I merged this branch into main in my fork, as I needed the feature, and I noticed a few issues:
- There are problematic `-}}` at a few places places in `_barman_object_store.tpl` in the handling of `destinationPath` and `$secretName`; it doesn't work as is, and moreover it complicates things, like having to do `{{ " destinationPath: \"s3://" }}` instead of just `destinationPath: "s3://`. 
- In `_backup.tpl` you pass a `secretPrefix` to `barmanObjectStoreConfig`, but in `_barman_object_store.tpl` you expect a `secretSuffix`.

itay-grudev:
@gpothier Have you fixed these in your fork? Can I take a look and/or steal your proposal?

gpothier:
Sure, I just committed it to my fork: https://github.com/gpothier/cnpg-charts/commit/fbdb794f384cfff84f583ce6b1ca62091cb0c2bf

I only fixed the yaml formatting stuff (using `}}` instead of `-}}`, as whitespace is part of YAML syntax). 

I don't know what your intent is regarding these prefix/suffix things so I did not touch that. 

I also added `quote` to `endpointURL` as I thought this was part of the problem; it wasn't, but I think it's better to quote in general anyway, and in particular for URLs that usually contain `:` signs that are not very safe to use in unquoted YAML values. I think it might be a good idea to have a look at other values that are not quoted. See <https://helm.sh/docs/chart_template_guide/functions_and_pipelines/>: 

> Let's start with a best practice: When injecting strings from the .Values object into the template, we ought to quote these strings.

gpothier:
Hi, have you been able to take a look at my suggested fix for this? Do you want me to create another PR with my changes?

itay-grudev:
I'm super overwhelmed with work right now, so yhea, I would really appreciate any help.

sxd:
@phisco can you give us a hand here? :D

itay-grudev:
@Cr4mble Thanks a lot! I would have totally missed that!

charts/cluster/templates/_barman_object_store.tpl
```diff
@@ -21,46 +21,49 @@
   {{- if empty .scope.destinationPath }}
   destinationPath: "s3://{{ required "You need to specify S3 bucket if destinationPath is not specified." .scope.s3.bucket }}{{ .scope.s3.path }}"
   {{- end }}
+  {{- $secretName := coalesce .scope.secret.name (printf "%s-%s-s3-creds" .chartFullname .secretPrefix) -}}
```
Cr4mble, line 24: There is a problem in rendering the chart locally if i use the s3 backup.
It should be
```suggestion
  {{- $secretName := coalesce .scope.secret.name (printf "%s-%s-s3-creds" .chartFullname .secretPrefix) }}
```

Cr4mble:
@itay-grudev can confirm that the backup to s3 is now working with the current state of the helm chart. 
The wal files are created in the specific s3 bucket.
I used the following values for testing (some are default values):
```
backups:
  enabled: true
  provider: s3
  secret:
    create: false
    name: "cloudnative-pg-s3-creds"
  s3:
    region: "eu-central-1"
    bucket: "testing-cloudnative-pg-cluster-helm-chart"
    path: ""
  scheduledBackups:
    -
      name: daily-backup
      schedule: "0 0 0 * * *"
      backupOwnerReference: self
      method: barmanObjectStore
```

itay-grudev:
@Cr4mble Thanks a lot.

8ball030:
Thanks a bunch for this guys. 
Really love the collaboration here :)

phisco (approved):

PR #240, opened 2024-04-01
Bug Fix: barmanObjectStore.serverName not initialized correctly

itay-grudev:
Bug Fix: barmanObjectStore.serverName not initialized correctly preventing recovery when the cluster name is different.


phisco (approved):

gpothier:
Hi, this change makes the `recovery-object_store` example obsolete: https://github.com/cloudnative-pg/charts/blob/0179d31d0784fa41a04bdf54fd47c9fe0691ea2c/charts/cluster/examples/recovery-object_store.yaml#L5

itay-grudev:
I need to fix that parameter to `clusterName` in the example. It was `clusterName` in the values and I wanted to call it `clusterName` instead of `serverName` everywhere in the chart is it makes more sense.

PR #247, opened 2024-04-04
feat(cluster): Make roles configurable

### LINKED ISSUE #201: Add support for declarative role management

gpothier:
This makes it possible to configure roles using the chart by adding an array in `cluster.roles` in `values.yaml`. The content of the array are taken as is and used as `.spec.managed.roles.`

This is a follow up to PR #226, which ended up with conflicts, so I started a new one.

In this new PR I signed off my changes, and I ran `make schema docs`.

The rationale for adding the `with` inside instead of outside `managed:` in `cluster.yaml` is that in case it is needed at some point to bring other things into `managed`, we would have to put an `if` with several conditions outside, and several `ifs` inside in order to avoid the empty element. I did try the chart without specifying roles to check if the empty `managed` was a problem, and there was no problem.

Closes #201

### DIFF
--- a/charts/cluster/templates/cluster.yaml
+++ b/charts/cluster/templates/cluster.yaml
@@ -57,6 +57,12 @@ spec:
       {{- toYaml . | nindent 6 }}
     {{ end }}
 
+  managed:
+    {{- with .Values.cluster.roles }}
+    roles:
+      {{- toYaml . | nindent 6 }}
+    {{ end }}
+
   monitoring:
     enablePodMonitor: {{ and .Values.cluster.monitoring.enabled .Values.cluster.monitoring.podMonitor.enabled }}
     {{- if not (empty .Values.cluster.monitoring.customQueries) }}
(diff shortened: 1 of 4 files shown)


charts/cluster/templates/cluster.yaml
```diff
@@ -57,6 +57,12 @@ spec:
       {{- toYaml . | nindent 6 }}
     {{ end }}
 
+  managed:
```
itay-grudev, line 60: Can you move the line above so the managed section is not included empty?

itay-grudev (changes requested):

itay-grudev:
Just one comment. Otherwise looks good.

charts/cluster/templates/cluster.yaml
gpothier: Sure I can do that, I just wonder if you have a specific disagreement with the rationale I provided for having the `{{ with ... }}` inside `managed`? 

If at some point there are additional elements in `managed` that you want to be able to express in the chart, you would have to do something like that to avoid the empty `managed`:

```
  {{- if or (.Values.cluster.roles) (.Values.cluster.someNewStuff) }}
  managed:
    {{- with .Values.cluster.roles }}
    roles:
      {{- toYaml . | nindent 6 }}
    {{ end }}
    {{- with .Values.cluster.someNewStuff }}
    roles:
      {{- toYaml . | nindent 6 }}
    {{ end }}
  {{ end }}
```

And AFAICT, having the empty `managed` item has no effect.

gpothier: Hi, any news on this? If you have a strong opinion on why to avoid rendering the empty `managed` element I can do as you suggest, I would just like to understand the rationale given the argument I provided.

itay-grudev: Your argument makes sense.

itay-grudev (approved):

PR #251, opened 2024-04-08
bugfix(cluster): Propagate inheritFromAzureAD to the barman configuration

Pionerd:
Currently the cluster chart allows setting inheritFromAzureAD, but this is not propagated into the configuration.


itay-grudev:
Have you tested this? I don't have access to such a setup, but it looks alright when checking the documentation.

Pionerd:
I tested the configuration with a separate `Cluster` resource and validated that a `helm template` from this branche gives the same output.

itay-grudev (approved):

PR #253, opened 2024-04-10
Update charts/cluster/README.md - fix Getting Started doc markdown link by escaping space

jessebot:
This add sandwiches the `Getting Started.md` file link in `<>` so that the space in the file is escaped, which will then render the link in the `README.md` when viewing on GitHub.

### DIFF
--- a/charts/cluster/README.md
+++ b/charts/cluster/README.md
@@ -51,7 +51,7 @@ helm upgrade --install cnpg \
 cnpg/cluster
 ```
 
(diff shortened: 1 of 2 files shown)


itay-grudev:
Please note that the README is auto-generated from [`README.md.gotmpl`](https://github.com/cloudnative-pg/charts/blob/main/charts/cluster/README.md.gotmpl) by running `make docs`. But even if you don't fix that I'll whenever we do the next release.

itay-grudev:
Thanks a lot!

itay-grudev (approved):

PR #286, opened 2024-05-21
fix init db to actually provision with a secret

8ball030:
helm.go:84: [debug] admission webhook "mcluster.cnpg.io" denied the request: json: cannot unmarshal string into Go struct field BootstrapInitDB.spec.bootstrap.initdb.secret of type v1.LocalObjectReference

The defazult values fail to provision,

In order to get the secret to work, you have to add the name key,.

### DIFF
--- a/charts/cluster/values.yaml
+++ b/charts/cluster/values.yaml
@@ -192,7 +192,8 @@ cluster:
   initdb: {}
     # database: app
     # owner: "" # Defaults to the database name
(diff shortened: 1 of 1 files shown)


8ball030:
I believe theere additional issues in in that when you try to use this configuration option, the new database is not correctly configured to use the secret specified.

Any guidance on this would be appreciated and i can update the pr to provide more clarity to others.

itay-grudev:
Can you share your values you are experiencing an issue with?

itay-grudev:
Closing in favor of: #239

PR #290, opened 2024-05-24
feat(cluster): add the possibility to use an existing secret for cloud credentials

### LINKED ISSUE #197: Add support for existing backup credentials

Cr4mble:
Our setup includes ArgoCD and SealedSecrets. As in the current helm chart the cloud credentials need to be in plaintext we can't commit this to git (GitOps).
That's why we adjusted the helm chart to accept an existing secret that we deployed earlier with ArgoCD and SealedSecrets.
We are currently using this for AWS s3 and the backup is working as expected.

We tried to adjust the helm chart as less as possible.

This PR should fix #197

### DIFF
--- a/charts/cluster/templates/_barman_object_store.tpl
+++ b/charts/cluster/templates/_barman_object_store.tpl
@@ -23,10 +23,10 @@
   {{- end }}
   s3Credentials:
     accessKeyId:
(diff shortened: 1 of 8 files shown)


itay-grudev:
Thank you for your contribution, but there is already a PR for this: #239. We also have other pending work that is built on top of it, so I'd rather keep that.

Would you be able to test that to help move the PR along?

PR #303, opened 2024-05-29
Add 'additionalEnvVars' to the chart

Stevenpc3:
Add ability to set additional ENV vars via ```additionalEnv``` value.

Array containing extra environment variables which can be templated. You can add this to an additional override file or example putting this in override.yaml
```
additionalEnv:
  - name: "{{ .Release.Namespace }}"
    value: "{{ .Release.Namespace }}"
  - name: MY_VAR
    value: mySpecialKey
  - name: 1234
    value: 5678
  - name: lottery
    value: 12345
  - name: bool
    value: false
  - name: empty
    value: ""
```

Then deploy with ``` helm upgrade -i cnpg -n namespace-a charts/cnpg/ -f override.yaml```

will generate (partial template below)

image

With these logs

image

### DIFF
--- a/charts/cloudnative-pg/Chart.yaml
+++ b/charts/cloudnative-pg/Chart.yaml
@@ -18,7 +18,7 @@ name: cloudnative-pg
 description: CloudNativePG Operator Helm Chart
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: "0.22.0"
+version: "0.21.2"
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning, they should reflect the version the application is using.
--- a/charts/cloudnative-pg/Chart.yaml
+++ b/charts/cloudnative-pg/Chart.yaml
@@ -18,7 +18,7 @@ name: cloudnative-pg
 description: CloudNativePG Operator Helm Chart
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: "0.21.2"
+version: "0.22.0"
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning, they should reflect the version the application is using.
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -66,6 +66,14 @@ config:
 # -- Additinal arguments to be added to the operator's args list.
 additionalArgs: []
 
+# -- Array containing extra environment variables which can be templated.
+# For example:
+#  - name: WATCH_NAMESPACE
+#    value: "{{ .Release.Namespace }}"
+#  - name: MY_VAR
+#    value: "mySpecialKey"
+additionalEnvVars: []
+
 serviceAccount:
   # -- Specifies whether the service account should be created.
   create: true
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -68,11 +68,11 @@ additionalArgs: []
 
 # -- Array containing extra environment variables which can be templated.
 # For example:
-#  - name: WATCH_NAMESPACE
-#    value: "{{ .Release.Namespace }}"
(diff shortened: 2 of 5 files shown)


charts/cloudnative-pg/Chart.yaml
```diff
@@ -18,7 +18,7 @@ name: cloudnative-pg
 description: CloudNativePG Operator Helm Chart
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: "0.21.2"
+version: "0.22.0"
```
itay-grudev, line 21: Could you drop this change. We bump the version number as part of our release process. Moreover, given that this is a non-breaking change, a minor release would suffice.

itay-grudev, line 21: Could you drop this change. We bump the version number as part of our release process. Moreover, given that this is a non-breaking change, a minor release would suffice.

charts/cloudnative-pg/values.yaml
```diff
@@ -66,6 +66,14 @@ config:
 # -- Additinal arguments to be added to the operator's args list.
 additionalArgs: []
 
+# -- Array containing extra environment variables which can be templated.
+# For example:
+#  - name: WATCH_NAMESPACE
+#    value: "{{ .Release.Namespace }}"
```
itay-grudev, line 72: Please remove this example. It implies this is somehow officially supported - whereas it's truly not. The core team said they will look into it, but for now it remains unsupported.

charts/cloudnative-pg/values.yaml
```diff
@@ -66,6 +66,14 @@ config:
...
+#  - name: MY_VAR
+#    value: "mySpecialKey"
+additionalEnvVars: []
```
itay-grudev, line 75: I believe the predominant convention is `additionalEnv`, and my colleagues love conventions.

itay-grudev (changes requested):
The `cloudnative-pg.extraEnv` doesn't looks like it has an issue. Does it work with multiple values at all?

Stevenpc3:
> The `cloudnative-pg.extraEnv` doesn't looks like right to me. Does it work with multiple values at all?

If you further review my description you can see where I use two values and then pass them in and then highlight the results in the configmap which shows both values passed in.

itay-grudev:
> > The `cloudnative-pg.extraEnv` doesn't looks like right to me. Does it work with multiple values at all?
> 
> Yes it does work with multiple values. If you further review my description you can see where I use two values and then pass them in and then highlight the results in the configmap which shows both values passed in.

I think the check whether the value is string only makes sense when you wrap your macro in a range - i.e.:
```yaml
{{- range $e := .Values.additionalEnv }}
{{- include "cloudnative-pg.extraEnv" (dict "value" $e "context" $ ) | nindent 8 }}
{{- end }}
```

Otherwise .value will always be an array

itay-grudev:
And then why don't you always do `tpl`. If there is no `{{` sequence there wouldn't be anything to replace.

Stevenpc3:
> > > The `cloudnative-pg.extraEnv` doesn't looks like right to me. Does it work with multiple values at all?
> > 
> > 
> > Yes it does work with multiple values. If you further review my description you can see where I use two values and then pass them in and then highlight the results in the configmap which shows both values passed in.
> 
> I think the check whether the value is string only makes sense when you wrap your macro in a range - i.e.:
> 
> ```yaml
> {{- range $e := .Values.additionalEnv }}
> {{- include "cloudnative-pg.extraEnv" (dict "value" $e "context" $ ) | nindent 8 }}
> {{- end }}
> ```
> 
> Otherwise .value will always be an array

Ranging over means I need to add my own "-" before name for list conversion.  Here is teh results of the template above and the override.yaml below.  You can see how range changed the items from a list to individial line outputs.

image

Stevenpc3:
> And then why don't you always do `tpl`. If there is no `{{` sequence there wouldn't be anything to replace.

Yeah this is good.  I made this change and it works well with how I testsed below.

image

Stevenpc3:
I simplified it and used the test cases of [] and the list above.  It appears good and not sure I can simplify any further.

itay-grudev (approved):

phisco (approved):
Awesome! Thanks @Stevenpc3 and @itay-grudev!

Issue #297, opened 2024-05-25
feat(cluster): Add support for TimescaleDB


itay-grudev:
After the following PRs get merged in the official TimescaleDB's HA docker container: 
* timescale/timescaledb-docker-ha/pull/442
* timescale/timescaledb-docker-ha/pull/459

and with the introduction of the `ImageCatalog` feature it's now possible to run TimescaleDB in CloudNativePG.

There is some code for that already added in the chart, but there are a lot of crucial components missing.

Depends on:
#276

PR #277, opened 2024-05-03
feat(cluster): Add ability to specify imageCatalogRef for Cluster resource

paraddise:
Issue - https://github.com/cloudnative-pg/charts/issues/276

Issue #276, opened 2024-05-02
feat(cluster): Add support for ImageCatalog


paraddise:
We already have cloudnative-pg 0.21.2 chart with operator version 1.23.1, in 1.23 added support or ImageCatalog, we need to specify this in cluster chart.

PR #306, opened 2024-05-30
feat( cluster ): cluster.imageCatalogRef; ImageCatalog support; TimescaleDB support

itay-grudev:
* Added support for `cluster.imageCatalogRef` (@paraddise)
* Added support for custom image catalogs via: `imageCatalog`
* Added support for the official TimescaleDB docker image via an embedded ImageCatalog

There is a bit of complication and checks, when using TimescaleDB, because the default `UID` and `GID` need to be set to `1000`, when using the [`timescaledb-ha`](https://github.com/timescale/timescaledb-docker-ha) docker image.

This PR adds the following new field `version` that allows you to select the major PostgreSQL version, as well as TimescaleDB and PostGIS versions:

```yaml
version:
  # -- PostgreSQL major version to use
  postgresql: "16"
  # -- If using TimescaleDB, specify the version
  timescaledb: "2.15"
  # -- If using PostGIS, specify the version
  postgis: "3.4"
```

P.S. The PR intentionally doesn't use the official `ClusterImageCatalog`s for backwards compatibility and instead sets `cluster.imageName`.

Closes: #276 
Closes: #297
Closes: #277


paraddise (approved):

jaredbischof (approved):
I've tested and validated the TimescaleDB deployment in five different Kubernetes environments.

charts/cluster/templates/_helpers.tpl
```diff
@@ -51,6 +51,20 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/part-of: cloudnative-pg
 {{- end }}
 
+{{/*
+Whether we need to use TimescaleDB defualts
```
phisco, line 55:
```suggestion
Whether we need to use TimescaleDB defaults
```

Issue #311, opened 2024-06-10
gkeEnvironment conflict?


rocket357:
https://github.com/cloudnative-pg/charts/blob/b378ad06aee0e9902eaaf8413784e5e3bb724861/charts/cluster/templates/_barman_object_store.tpl#L64-L67

I'm setting up a test environment in gke, and I've set the following in values:

```
backups:
  enabled: true
  provider: google
  google:
    gkeEnvironment: true
    bucket: my-test-bucket
    path: my-test-path
```

Via the linked template above, I see the following:

```
        googleCredentials:
        gkeEnvironment: false
        applicationCredentials:
          name: cnpg-database-cluster-backup-google-creds
          key: APPLICATION_CREDENTIALS
```

With this configuration, I'm seeing the following in helm output:

```
COMBINED OUTPUT:
  Error: UPGRADE FAILED: release cloudnative-pg-database-cluster failed, and has been rolled back due to atomic being set: cannot patch "cloudnative-pg-database-cluster" with kind Cluster: admission webhook "vcluster.cnpg.io" denied the request: Cluster.cluster.cnpg.io "cloudnative-pg-database-cluster" is invalid: spec.backupConfiguration.googleCredentials: Invalid value: v1.GoogleCredentials{ApplicationCredentials:(*v1.SecretKeySelector)(0xc001e0ada0), GKEEnvironment:true}: if gkeEnvironment is true, secret with credentials must not be provided
```

So it looks like the [docs](https://cloudnative-pg.io/documentation/1.18/api_reference/#googlecredentials) are not entirely accurate (specifically: "GoogleCredentials is the type for the Google Cloud Storage credentials. This needs to be specified even if we run inside a GKE environment.")

The operator I'm running is cloudnative-pg latest (0.21.4).  Is there additional configuration I've overlooked?

jsilvela:
Hi @rocket357 , I'm reviewing the documentation and the webhook code, and I think the documentation is not wrong (though it could stand extra detail).
The problem reported by the webhook is bona fide:
> if gkeEnvironment is true, secret with credentials must not be provided

Now, looking at your case, you've supplied `gkeEnvironment: false` plus credentials, so the warning message would seem out of place.
But, in the tpl file I see `gkeEnvironment: {{ .scope.google.gkeEnvironment }}`.

And the value parsed shows the GKE is true.
```
[...]

PR #312, opened 2024-06-10
Add walStorage capability and remove gkeEnvironment conflict in cluster chart.

rocket357:
The operator supports walStorage, but the chart does not (yet).  This PR fixes that.

In addition, this PR addresses issue https://github.com/cloudnative-pg/charts/issues/311 so gkeEnvironment can be set without the operator admission webhook rejecting the update.

### DIFF
--- a/charts/cluster/templates/_barman_object_store.tpl
+++ b/charts/cluster/templates/_barman_object_store.tpl
@@ -62,8 +62,10 @@
   {{- $secretName := coalesce .scope.secret.name (printf "%s-%s-google-creds" .chartFullname .secretPrefix) }}
   googleCredentials:
     gkeEnvironment: {{ .scope.google.gkeEnvironment }}
+{{- if not .scope.google.gkeEnvironment }}
     applicationCredentials:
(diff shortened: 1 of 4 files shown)


rocket357:
I'm working on an additional chart issue (https://github.com/cloudnative-pg/charts/issues/311) where having gkeEnvironment set to true causes the admission webhook in the operator to reject updates (due to applicationCredentials being also set).  I've rolled that change in here as well.  If it's preferred for me to break these out into separate PRs, please let me know!

itay-grudev:
@rocket357 Not everyone would approve, but the PR is so tiny - I'm fine with it as is. Just make sure both issues are mentioned in the PR description.

itay-grudev (approved):

itay-grudev (changes requested):
Can you add the new parameter and a documentation line in the `values.yaml`?

rocket357:
Requested changes have been added.  Please let me know if there's anything else that needs to be updated!

itay-grudev (approved):

PR #321, opened 2024-06-18
cluster: Add possibility to configure custom pg_hba

### LINKED ISSUE #308: Setting up pg_hba not working
Hello everyone,
I'm trying to set pg_hba conf but the helm values don't seem to work.
What I've tried so far is:

```
...
  postgresql:
    parameters:
      shared_buffers: 256MB
      pg_stat_statements.max: '10000'
      pg_stat_statements.track: all
      auto_explain.log_min_duration: '10s'
    pg_hba: "host all all all password"
```

or 

```
  postgresql:
      shared_buffers: 256MB
      pg_stat_statements.max: '10000'
      pg_stat_statements.track: all
      auto_explain.log_min_duration: '10s'
      pg_hba: "host all all all password"
```

But i'm getting this error:

`LOG:  unrecognized configuration parameter "pg_hba" in file "/var/lib/postgresql/data/pgdata/custom.conf" line 20"`

mjtrangoni:
Hi,

This is a **BREAKING CHANGE**, and fixes #308. 

IMHO, we should go with this kind of fixes now and taking into account that **postgresql** has a lot of configurations defined [here](https://cloudnative-pg.io/documentation/current/cloudnative-pg.v1/#postgresql-cnpg-io-v1-PostgresConfiguration), not only *parameters*, this fixes the way it will be handled.

Thank you!

PS: I had to remove one trailing space for the the CI to success, but in a separate commit so that you can remove it easily


itay-grudev:
I agree with this change. The advantage of running with version `< 0.x,x` is that breaking changes can happen while evaluating the initial prototype in the wild.

I will release it as `0.1.0` and add the relevant changelog entry.

itay-grudev:
@mjtrangoni Could you please run `make schema`?

mjtrangoni:
Hi @itay-grudev, 

Thanks for your prompt reply. It seems I can't run it on my end, because the *schema-gen* does not support darwin, and is also not maintained anymore. Any hints on how to override this?

See,
```zsh
$make schema
Error: unknown command "schema-gen" for "helm"
Run 'helm --help' for usage.
Error: unknown command "schema-gen" for "helm"
Run 'helm --help' for usage.
$ helm plugin install https://github.com/karuppiah7890/helm-schema-gen.git
karuppiah7890/helm-schema-gen crit platform darwin/arm64 is not supported.  Make sure this script is up-to-date and file request at https://github.com/karuppiah7890/helm-schema-gen/issues/new
Error: plugin install hook for "schema-gen" exited with error
```

mjtrangoni:
@itay-grudev tests are green now. pg_hba should be an array and not an object. Could you fix the README.md as well?

itay-grudev:
I was just going through the documentation and found, that we've documented the `postgresql:` as a proxy to the [full configuration](https://cloudnative-pg.io/documentation/current/cloudnative-pg.v1/#postgresql-cnpg-io-v1-PostgresConfiguration).

So this means that this is no longer a simply a braking change, but also a bug fix.

itay-grudev (approved):

mjtrangoni:
@itay-grudev is there any release date for this?

itay-grudev:
@mjtrangoni I apologise about the delay. I wanted to merge some other work first and then this, as this would require a minor patch due to breaking changes. If I get everything reviewed next week, I'll merge it.

itay-grudev:
P.S. The moment I update this branch in a moment, the test should fail due to the test suite that I just added to the chart. The test will need to be updated, but I'll take care of that. I have edit access to your branch.

PR #335, opened 2024-07-24
Fix VolumeSnapshot backup capability

### LINKED ISSUE #334: VolumeSnapshot doesn't work
* Missing [VolumeSnapshotConfiguration](https://cloudnative-pg.io/documentation/1.23/cloudnative-pg.v1/#postgresql-cnpg-io-v1-VolumeSnapshotConfiguration) block on the Cluster resource. This prevents the `Backup` from working correctly when we set `backups.scheduledBackups[*].method` to `volumeSnapshot`
* The chart requires azure, google, or s3 options even when we choose to use volumeSnapshot as a provider

cam-at-tactiq:
Hi this is my first contribution so please let me know if I have missed anything.

This PR fixes two issues with VolumeSnapshots for backups:
* Missing [VolumeSnapshotConfiguration](https://cloudnative-pg.io/documentation/1.23/cloudnative-pg.v1/#postgresql-cnpg-io-v1-VolumeSnapshotConfiguration) block on the Cluster resource
* Prevents the chart from complaining about missing azure, google, or s3 options when we choose to use volumeSnapshot as a provider
* Add support for recovery from a VolumeSnapshot

This is achieved by establishing 'volumeSnapshot' as a `backups.provider`.

There are no breaking changes.

Fixes #334


itay-grudev:
This looks pretty good! 

Would you have the time to add the [Recovery from VolumeSnapshots](https://cloudnative-pg.io/documentation/1.23/recovery/#recovery-from-volumesnapshot-objects) functionality as well?

It wouldn't want to add this as a backup method without adding a recovery path, as that could potentially make it more difficult to recover from.

cam-at-tactiq:
Thanks @itay-grudev! I've added support for recovery from snapshots and tested this in a GKE environment. The recovery [config docs](https://cloudnative-pg.io/documentation/1.23/cloudnative-pg.v1/#postgresql-cnpg-io-v1-BootstrapRecovery) also mention being able to use a PVC as a recovery mechanism with the same block on the cluster, but that failed when I tested it, otherwise I would have added that too. It might be a nice to have when that capability stabilises in the operator.

Please let me know if there's anything else that could make this better.

cam-at-tactiq:
Hi @itay-grudev please let me know what blockers if any there are to merging, we are really keen to start using this when it is released :)

dragoangel:
I not think that removing capability of chose between S3 and volumeSnapshot is good idea actually. Provider allow to utilize both S3 and Snapshots, why we in Helm should cut off this capabilities, I see this option as: https://github.com/cloudnative-pg/charts/pull/359

cam-at-tactiq:
> I not think force user to decide between s3 and snapshot is good idea actually. Provider allow to utilize both S3 and snapshots, why we in Helm should cut off this capabilities, I see this option as: #359

Hi @dragoangel, I like your PR a lot. Whatever the maintainers prefer I am happy with, I just really want some volume snapshot capability. I've posted a review on your PR with two suggested changes. @itay-grudev please feel free to close this PR if you are going to merge #359 .

cam-at-tactiq:
The solution here is offered freely, but clearly isn't wanted by maintainers at this point in time.

PR #343, opened 2024-07-30
added pg_ident and pg_hba to cluster.yaml as values overrides

nightcapneal:
added pg_ident and pg_hba to cluster.yaml as values overrides


itay-grudev:
I'll change the API to make all of these available under the `cluster.postgres.*`. I'm closing this for now.

Stevenpc3:
Is there an ETA on when to expect that update?

Issue #317, opened 2024-06-12
Documentation for endpointCA feature #230


mboncalo:
Hi,

I've been trying the CNPG operator with cluster helm charts and I couldn't make the S3 backup work.
We have an Minio S3 CRD provided by our cloud provider (Gardner) and a secret with all S3 bucket details including the endpointCABundle.
I tried creating a new secret:
endpointCA: 

    # -- Creates a secret with the given value if true, otherwise uses an existing secret.
    create: true
    name: "bucket-postgres-service-1"
    key: "s3_ca_bundle"
    value: "IyBrdWJlLX..." # CA bundle value from bucket secret

Use the existing secret:

endpointCA: 

    # -- Creates a secret with the given value if true, otherwise uses an existing secret.
    create: false
    name: "s3-bucket-postgres-service-1"
    key: "s3_ca_bundle"
    value: ""

destinationPath: ""
  # -- One of `s3`, `azure` or `google`
  provider: s3
  s3:
    region: ""      **#tried "default" and "us-east-1" as well**
    bucket: "5437053e567993e87f73a0cd1aafb5b04f1b85334d32a534acb8a1b8"
    path: "/"
    accessKey: "<key>
    secretKey: "<secret>"
  azure:
    path: "/"
    connectionString: ""
    storageAccount: ""
    storageKey: ""
    storageSasToken: ""
    containerName: ""
    serviceName: blob
    inheritFromAzureAD: false
  google:
    path: "/"
    bucket: ""
    gkeEnvironment: false
    applicationCredentials: ""
  secret:
    # -- Whether to create a secret for the backup credentials
    create: true
    # -- Name of the backup credentials secret
    name: "service-1-backup"

scheduledBackups:
    -
      # -- Scheduled backup name
      name: daily-backup
      # -- Schedule in cron format
      schedule: "0 */5 * * * *"
      # -- Backup owner reference
      backupOwnerReference: self
      # -- Backup method, can be `barmanObjectStore` (default) or `volumeSnapshot`
      method: barmanObjectStore

  # -- Retention policy for backups
  retentionPolicy: "30d"

Operator says in the logs that it's doing backups:
`
[...]

PR #230, opened 2024-03-26
Add config value endpointCA for private S3 such as MinIO

PseudoResonance:
This PR adds a chart values to `endpointCA` in both the backup and recovery sections, allowing a user to specify a private CA. The value is stored in a secret, which can either be created manually, or will be created automatically if the `create` option is true.

Resolves #229

PR #341, opened 2024-07-30
Kyverno Chainsaw Test Suite

itay-grudev:
This implements a base set of tests for the following areas using [Kyverno Chainsaw](https://kyverno.github.io/chainsaw/latest/):

* Backup/restore from a `Backup` using MinIO
* Backup/restore from Object Store using MinIO
* Backup/restore from Object Store with PITR (time target) using MinIO
* Monitoring settings
* Pooler (to maintain the same functionality of the existing test I am replacing)

I am going to stop here, but the following further tests could be addressed in a separate PR:

* [Azurite](https://github.com/Azure/Azurite) tests for Azure Data Store compatibility
* `type: postgis` tests

This PR includes the fix in #345 for the MinIO backup/restore tests to work.

Closes #294

PR #345, opened 2024-08-01
Bug Fix: Endpoint CA settings not passed correctly to the cluster

itay-grudev:
This fixes an issue outlined in #317, that was introduced in #230, where the correct values of `endpointCA` are not passed to the cluster.


itay-grudev:
This was merged in #341

PR #355, opened 2024-08-23
feat(cloudnative-pg): add support for CRD annotations

afarbos:
Adding support for CRD annotations can be really usefull especially with deployment services like argocd, spinnaker or others.
For example, in my case I would like to set this annotations due to the size of the CRDs:
```
    annotations:
      argocd.argoproj.io/sync-options: ServerSideApply=true
```
Ref docs: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#server-side-apply

### DIFF
--- a/charts/cloudnative-pg/templates/crds/crds.yaml
+++ b/charts/cloudnative-pg/templates/crds/crds.yaml
@@ -5,6 +5,9 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.15.0
     helm.sh/resource-policy: keep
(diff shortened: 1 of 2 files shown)


itay-grudev:
As part of the operator release process we generate the specs and we just put the generated CRDs in the chart so any changes would be overwritten.

Furthermore, CRDs should not be templated by Helm. It's bad practice and disabled by default. I will in fact move the `crds` outside the `templates` directory as per Helm recommendation as otherwise they are not always applied before other resources that depend on them.

Issue #354, opened 2024-08-23
Add replication slots highAvailability settings


dragoangel:
Section for replication slots is missing in cluster chart https://cloudnative-pg.io/documentation/1.20/replication/#replication-slots-for-high-availability

PR #356, opened 2024-08-26
feat(cluster): Add replication slots configuration

dragoangel:
implements: https://github.com/cloudnative-pg/charts/issues/354


charts/cluster/values.yaml
```diff
@@ -153,6 +153,15 @@ cluster:
...
   superuserSecret: ""
 
+  # -- This feature automatically manages physical replication slots for each hot standby replica in
+  # the High Availability cluster, both in the primary and the standby.
+  # See: https://cloudnative-pg.io/documentation/current/replication/#replication-slots-for-high-availability
+  replicationSlots:
+    highAvailability:
+      enabled: true
+      slotPrefix: _cnpg_
+      updateInterval: "30"
```
itay-grudev, line 163: Can you replace this with:

```
  replicationSlots: {}
   # highAvailability:
   #   enabled: true
   #   slotPrefix: _cnpg_
   #   updateInterval: "30"
```

and then re-run `make schema docs`.

Otherwise the generated schema is incorrect and would not allow otherwise valid configurations. This way it would only validate whether it's passed an object.

itay-grudev (changes requested):

charts/cluster/values.yaml
dragoangel: Sorry, I not get a bit what you mean. This will be this way, what else can be set outside of this schema?

itay-grudev: Refer to the API reference here: https://cloudnative-pg.io/documentation/current/cloudnative-pg.v1/#postgresql-cnpg-io-v1-ReplicationSlotsConfiguration

itay-grudev: The problem is that our schema is auto generated based on values, so it will assume the structure you provided, which is incomplete. See `make schema`.

dragoangel: I found and fixed issue, thanks

markediez:
@itay-grudev / @dragoangel what needs to be done to get this merged?

PR #357, opened 2024-08-26
feat(cluster): Allow to deploy multiple poolers

dragoangel:
implements: https://github.com/cloudnative-pg/charts/issues/360

### DIFF
--- a/charts/cluster/templates/pooler.yaml
+++ b/charts/cluster/templates/pooler.yaml
@@ -1,21 +1,22 @@
-{{ if .Values.pooler.enabled }}
+{{ range .Values.poolers -}}
+---
 apiVersion: postgresql.cnpg.io/v1
 kind: Pooler
 metadata:
-  name: {{ include "cluster.fullname" . }}-pooler-rw
+  name: {{ include "cluster.fullname" $  }}-pooler-{{ .name }}
 spec:
   cluster:
-    name: {{ include "cluster.fullname" . }}
-  instances: {{ .Values.pooler.instances }}
-  type: {{ .Values.pooler.type }}
+    name: {{ include "cluster.fullname" $ }}
+  instances: {{ .instances }}
+  type: {{ .type }}
   pgbouncer:
-    poolMode: {{ .Values.pooler.poolMode }}
+    poolMode: {{ .poolMode }}
     parameters:
-      {{- .Values.pooler.parameters | toYaml | nindent 6 }}
+      {{- .parameters | toYaml | nindent 6 }}
   monitoring:
-    enablePodMonitor: {{ and .Values.pooler.monitoring.enabled .Values.pooler.monitoring.podMonitor.enabled }}
-  {{- with .Values.pooler.template }}
+    enablePodMonitor: {{ and .monitoring.enabled .monitoring.podMonitor.enabled }}
+  {{- with .template }}
   template:
     {{- . | toYaml | nindent 4 }}
   {{- end }}
-{{ end }}
+{{ end -}}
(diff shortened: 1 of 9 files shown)


itay-grudev:
Can you sync with me on the CloudNativePG slack?

I can't review these until the weekend, but I can give you guidance, so it's as easy as possible.

dragoangel:
Hi @itay-grudev, sure 😊

dragoangel:
In short about this PR: as it contains **breaking change** it will be merged a bit later.

charts/cluster/templates/pooler.yaml
```diff
@@ -1,21 +1,25 @@
-{{ if .Values.pooler.enabled }}
+{{ if gt (len .Values.poolers) 0 }}
```
itay-grudev, line 1: Why do we need the outside check? The `range` should be sufficient.

charts/cluster/templates/pooler.yaml
```diff
@@ -1,21 +1,25 @@
...
+{{ $context := . -}}
```
itay-grudev, line 2: No need to cache the context. Inside the `range` you can access the top context by using `$`. Example: `$.Values`

charts/cluster/test/monitoring/01-monitoring_cluster.yaml
```diff
@@ -18,8 +18,9 @@ cluster:
...
-  enabled: true
-  instances: 1
-  monitoring:
-    enabled: true
+poolers:
+  - name: rw
+    type: rw
+    instances: 1
+    monitoring:
+      enabled: true
```
itay-grudev, line 26: Add at least two poolers to the test to verify the code above generates the YAML correctly.

itay-grudev (changes requested):

charts/cluster/test/monitoring/01-monitoring_cluster.yaml
dragoangel: I added 2 poolers into poolers test, was expect it was enough, but sure, can add here too.

itay-grudev: Oh! I didn't realize that. But this is also good. Thanks.

itay-grudev (approved):

charts/cluster/templates/pooler.yaml
dragoangel: To note: I taken it from scheduled backups, I think we should change it there as well?

itay-grudev:
@dragoangel You're missing the monitoring section in the per-pooler configuration, hence the error. This wasn't a problem before since all the properties were always set. The best way to fix that would be an extensive use of `default` I suppose.

P.S. Because the old property is `pooler` and the new one `poolers`, this can be made backwards compatible.

dragoangel:
@itay-grudev you mean in auto tests?

dragoangel:
Hi @itay-grudev I pushed proper label example and tests.

I see 3 options:
1. leave as is
2. move `.Values.cluster.monitoring` and `.Values.pooler.monitoring` to `.Values.monitoring`, and leave podMonitor relablings under `cluster` and each `pooler` if they need to get this settings.
3. remove `.Values.poolers.[0].monitoring.enabled` and `.Values.poolers.[0].monitoring.podMonitor.enabled` and take this info from `cluster`, while still looks for `relablings` under specific pooler. It will assume - if we create pod monitor for cluster we want to have pod monitor for pooler too, but this not always can be true?

For me all 3 options are okay, but 1 and 2 looks better.

itay-grudev (approved):

itay-grudev:
@dragoangel FYI: I went with option 1 as it was the simplest.

PR #358, opened 2024-08-26
feat(cluster): Extend monitoring options with relabelings

dragoangel:
Allow to customize resources as supported at https://cloudnative-pg.io/documentation/1.20/cloudnative-pg.v1/#postgresql-cnpg-io-v1-PoolerMonitoringConfiguration


dragoangel:
Added `customQueriesSecret` as default list of secrets, so users who will use this chart as dependency can render their own secrets as they like (for example via `vault-secret-operator`), and have capability to point current chart to them. Current `customQueries` use `ConfigMap` and it will cover most of use cases, but in rare cases secrets also can be useful.

itay-grudev (approved):

PR #359, opened 2024-08-26
feat(cluster): Add volume snapshots settings

dragoangel:
This PR will cover [Configuration of Volume Snapshot backups](https://cloudnative-pg.io/documentation/current/backup_volumesnapshot/#how-to-configure-volume-snapshot-backups).

I'm aware of `chainsaw-test.yaml` but without knowing `className` it's impressible to test this functionality and of course without CSI that supports such capabilities in the cluster.


dragoangel:
@itay-grudev I also have strong feeling that better put all cloud related settings under dedicated level, so everything starting from `provider` (224 line) and ending `data.jobs` (279 line) will be placed under `objectStorage` section. What do you think?

charts/cluster/templates/_helpers.tpl
```diff
@@ -68,3 +68,18 @@ If a custom imageName is available, use it, otherwise use the defaults based on
...
+{{- define "cluster.cloudBackupsEnabled" -}}
+{{- $cloudProviders := list "s3" "azure" "google" -}}
+{{- and .Values.backups.provider (has .Values.backups.provider $cloudProviders) -}}
+{{- end }}
+
+{{/*
+Backups enabled
+*/}}
+{{- define "cluster.backupsEnabled" -}}
+{{- or (eq (include "cluster.cloudBackupsEnabled" .) "true") .Values.backups.volumeSnapshot.className -}}
```
cam-at-tactiq, line 84:
```suggestion
{{- or (eq (include "cluster.cloudBackupsEnabled" .) "true") (not (empty .Values.backups.volumeSnapshot.className )) -}}
```

charts/cluster/templates/_backup.tpl
```diff
@@ -15,5 +16,12 @@ backup:
 
     {{- $d := dict "chartFullname" (include "cluster.fullname" .) "scope" .Values.backups "secretPrefix" "backup" }}
     {{- include "cluster.barmanObjectStoreConfig" $d | nindent 2 }}
+  {{- end }}
+  {{- if .Values.backups.volumeSnapshot.className }}
```
cam-at-tactiq, line 20:
```suggestion
  {{- if (not (empty .Values.backups.volumeSnapshot.className )) }}
```

cam-at-tactiq (changes requested):
I like this solution a lot, this addresses issue #334 differently than #335 (my own PR). I don't mind whichever solution the maintainers prefer, so long as we get this capability :)

I tested this today and it worked perfectly with the two attached changes.

This PR needs
* update to values in `README.md`
* an example values file for backup and restore

dragoangel:
> This PR needs
> * update to values in `README.md`
> * an example values file for backup and restore

Hi @cam-at-tactiq , yes I know, yesterday I was created couple of functional PRs and almost in every of them need to do it, thanks for heads up. I will add them today, yesterday was too much exchasted.

charts/cluster/templates/_helpers.tpl
dragoangel: As far as I know string is false when empty, what this change improves?

cam-at-tactiq: when I tested this locally and set my preferred classname it didn't evaluate to true as it should have.

dragoangel: Will check thanks, and apply your suggestion

dragoangel: Yeah, you right, it should be done like you proposed here.

dragoangel:
> @itay-grudev I also have strong feeling that better put all cloud related settings under dedicated level, so everything starting from `provider` (224 line) and ending `data.jobs` (279 line) will be placed under `objectStorage` section. What do you think?

I decided to move everything under dedicated section, it greatly improves readability, plus having `providerSettings` section allows to dynamically check which `provider` is supported.

dragoangel:
Hi @cam-at-tactiq, if you have some time - could you please test my PR? From my tests everything works good, but second-third view wouldn't be out of place. Thank you in advance.

Hi @itay-grudev, I not done tests via `chainsaw` stuff, they require some knowledge and secrets about infra it will be run on. If you not mind can you please add them?
I provided all necessary examples, so I hope there should be no issues.

If you have any questions or propositions about this PR, do not hesitate ping me in Slack 😉. I understand it is big, and will require quite time to review, and yes it is **BREAKING CHANGE**

rc-networks:
Looks like this issue is stale for more than 2months now.
I've been waiting for the external database support so I can start migrating old databases to CNPG.

What seems to be blocking the merge?

cam-at-tactiq:
No idea tbh, we have two decent PRs that handle this issue but neither have gained traction with maintainers :/

PR #362, opened 2024-08-28
fix(cluster): avoid invalid type for initdb secret

### LINKED ISSUE #361: [BUG] cluster chart uses incorrect type for initdb secret
Hi,

when using the default values as a template for the deployment of a new cluster, I have noticed that the secret reference in `cluster.initdb.secret` should contain a `LocalReference`, which is an object with the `name` key, rather than just a string.

The problematic section contains:
```yaml
cluster:
  ...
  initdb: {}
    ...
    # This should be a `LocalReference` instead.
    # secret: "" # Name of the secret containing the initial credentials for the owner of the user database. If empty a new secret will be created from scratch
...
```

> **Note:** I will submit a PR for this shortly.

nilfr:
This fixes #361.

### DIFF
--- a/charts/cluster/values.yaml
+++ b/charts/cluster/values.yaml
@@ -203,8 +203,12 @@ cluster:
     # owner: "" # Defaults to the database name
     # secret:
     #   name: "" # Name of the secret containing the initial credentials for the owner of the user database. If empty a new secret will be created from scratch
+    # options: []
+    # encoding: UTF8
     # postInitSQL:
     #   - CREATE EXTENSION IF NOT EXISTS vector;
+    # postInitApplicationSQL: []
+    # postInitTemplateSQL: []
 
   additionalLabels: {}
   annotations: {}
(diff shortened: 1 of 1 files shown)


charts/cluster/values.schema.json
```diff
@@ -197,7 +197,29 @@
...
+                                }
+                            }
+                        },
+                        "postInitSQL": {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            }
+                        }
+                    }
```
itay-grudev, line 222: This bit isn't correct as there are many more options possible. Can you remove the changes here and leave the loose validation instead?

itay-grudev (changes requested):

charts/cluster/values.schema.json
nilfr: Sure!

itay-grudev (approved):

Issue #366, opened 2024-08-29
WAL Storage cannot be disabled in cluster-v0.0.10


brinnjoyce:
https://github.com/cloudnative-pg/charts/pull/312/commits/4d20f4f0f13ff99525222d04a62b473c9c51c9ab added a default value for walStorage to `values.yaml` and https://github.com/cloudnative-pg/charts/pull/312/commits/274c40e06275e185cc673860d7b30d5c9b6912c8 adds the walStorage config to the cluster manifest if `.Values.cluster.walStorage` is defined - i.e. walStorage is forced to enabled in cluster-v0.0.10 and cannot be disabled.

pschichtel:
I think a .enabled option should be added there that default's to false.

gecube:
@pschichtel Hi! I reported in [another repo](https://github.com/cloudnative-pg/cloudnative-pg/issues/5891) (it's my mistake, I should report here) that this change broke the behaviour. Fortunately, I was able to fix easily `values.yaml` but I would not recommend to do the same in the future.

itay-grudev:
@gecube There was indeed a typo. It was fixed and released in [`v0.0.11`](https://github.com/cloudnative-pg/charts/releases/tag/cluster-v0.0.11).

itay-grudev:
@gecube Regarding disabling it after the cluster was created with wallStorage - that's not supported by the operator. There are advantages in keeping it that way - but if you want to get rid of it - I'd suggest re-creating the cluster.

You can refer to the operator documentation [here](https://cloudnative-pg.io/documentation/current/storage/#volume-for-wal).

PR #367, opened 2024-08-29
Revert to walStorage disabled by default and add enabled flag to enable

brinnjoyce:
Fixes https://github.com/cloudnative-pg/charts/issues/366

Upgrading from cluster-v0.0.9 to cluster-v0.0.10 forces walStorage to enabled with no way to disable it.

This reverts to walStorage not being configured as in cluster-v0.0.9 unless `.Values.cluster.walStorage.enabled` is set to true.

### DIFF
--- a/charts/cluster/templates/cluster.yaml
+++ b/charts/cluster/templates/cluster.yaml
@@ -24,7 +24,7 @@ spec:
   storage:
     size: {{ .Values.cluster.storage.size }}
     storageClass: {{ .Values.cluster.storage.storageClass }}
-{{- if .Values.cluster.walStorage }}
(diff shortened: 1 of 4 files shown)


itay-grudev (approved):

itay-grudev:
Can you add your new field to the `Install the non-default configuration cluster` test so it passes?

brinnjoyce:
Coming back to this the next day I think it's cleaner to just set walStorage to an empty map in values.yaml so I've created https://github.com/cloudnative-pg/charts/pull/370 to do that.

I've added the parameter to the test in this PR if you prefer the enabled flag though.

PR #369, opened 2024-08-30
add PGDBNAME to database-ping-test

### LINKED ISSUE #368: Database ping test fails when the username and dbname is different in initdb
Hi everyone,

The database-ping-test fails when i set the value of the database different from the owner in the initdb configuration. 

Following is my configuration for cluster chart, I am using kustomize to deploy helm

```yaml
helmCharts:
  - name: cluster
    version: 0.0.10
    releaseName: psql-gitea
    repo: https://cloudnative-pg.github.io/charts
    namespace: homelab
    valuesInline:
      cluster:
        instances: 1
        initdb:
          database: giteadb
          owner: gitea
```

And this results in the following error in the ping-test pod:

```
psql: error: connection to server at "psql-gitea-cluster-rw.homelab.svc.cluster.local" (10.43.64.140), port 5432 failed: FATAL:  database "gitea" does not exist
```

As you can see it tries to connect to `gitea` database, even though we specified it as `giteadb`
The connection string and pod configuration:
```yaml
# psql-gitea-cluster-ping-test pod 
spec:
  containers:
  - args:
    - -c
    - apk add postgresql-client && psql "postgresql://$PGUSER:$PGPASS@psql-gitea-cluster-rw.homelab.svc.cluster.local:5432"
      -c 'SELECT 1'
    command:
    - sh
    env:
    - name: PGUSER
      valueFrom:
        secretKeyRef:
          key: username
          name: psql-gitea-cluster-app
    - name: PGPASS
      valueFrom:
        secretKeyRef:
          key: password
          name: psql-gitea-cluster-app
    image: alpine:3.17
```
When I manually try to connect with the giteadb, i can succesfully create a connection:

```
➜ kubectl exec -it psql-gitea-cluster-1 -c postgres -- psql 'postgresql://gitea:<AUTO_GENERATED_PASSWORD>@psql-gitea-cluster-rw.homelab:5432/giteadb' -c 'SELECT 1'
 ?column? 
----------
        1
(1 row)

``` 

So the problem is, in the test-pod there is no db specified so it (by default) tries to connect to a db with the same name as the user. As there is no db created as such, it fails.

UgurcanAkkok:
Fix #368 by adding `$PGDBNAME` env var to database-ping-test. 

Note that if the user uses a custom secret in the initdb, they will also need to add the dbname to the secret. 

I am happy to help in case of any missing changes for the fix.

### DIFF
--- a/charts/cluster/templates/tests/ping.yaml
+++ b/charts/cluster/templates/tests/ping.yaml
@@ -30,8 +30,13 @@ spec:
                 secretKeyRef:
                   name: {{ include "cluster.fullname" . }}-app
                   key: password
+            - name: PGDBNAME
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "cluster.fullname" . }}-app
+                  key: dbname
           args:
             - "-c"
             - >-
               apk add postgresql-client &&
-              psql "postgresql://$PGUSER:$PGPASS@{{ include "cluster.fullname" . }}-rw.{{ .Release.Namespace }}.svc.cluster.local:5432" -c 'SELECT 1'
+              psql "postgresql://$PGUSER:$PGPASS@{{ include "cluster.fullname" . }}-rw.{{ .Release.Namespace }}.svc.cluster.local:5432/$PGDBNAME" -c 'SELECT 1'
--- a/charts/cluster/templates/tests/ping.yaml
+++ b/charts/cluster/templates/tests/ping.yaml
@@ -35,8 +35,9 @@ spec:
                 secretKeyRef:
                   name: {{ include "cluster.fullname" . }}-app
                   key: dbname
+                  optional: true
           args:
             - "-c"
             - >-
               apk add postgresql-client &&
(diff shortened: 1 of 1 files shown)


charts/cluster/templates/tests/ping.yaml
```diff
@@ -30,8 +30,13 @@ spec:
                 secretKeyRef:
                   name: {{ include "cluster.fullname" . }}-app
                   key: password
+            - name: PGDBNAME
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "cluster.fullname" . }}-app
```
phisco, line 36: Not sure we can assume the database is always called this way 🤔

UgurcanAkkok: The secret name assumption is already made in the ping.yaml. If you have suggestions to make it more dynamic, I am happy to try to fix it. 

However, the new assumption is for users to add the `dbname` secret key. I am going to fix it so that it will be optional. Note that if the user left the `initdb.secret` value empty, the `dbname` is already set.

itay-grudev: This comes from the operator @phisco. It's always `app` AFAIK, even if you change `dbname`.

itay-grudev (approved):

PR #370, opened 2024-08-30
Change walStorage to an empty map in values.yaml

brinnjoyce:
Fixes https://github.com/cloudnative-pg/charts/issues/366
Replaces https://github.com/cloudnative-pg/charts/pull/367

This makes it optional so walStorage isn't forced to be included in the Cluster manifest


itay-grudev:
I kind if like your original solution with the `enabled` flag. I'll choose that instead. Thanks for the alternative though.

PR #376, opened 2024-09-03
ParadeDB support

itay-grudev:
Adds support for running a ParadeDB cluster using CloudNativePG. This PR makes the necessary changes needed for ParadeDB and adds a backup/restore test suite.


philippemnoel:
Hi folks. Phil here from @paradedb. We're very excited for this PR. Thank you for maintaining CloudNativePG. A large fraction of our enterprise users adopt us via CloudNativePG and being in the official charts would really help drive adoption even further.

itay-grudev:
The upstream chart has diverged enough that ParadeDB plans to maintain their own chart separately. For ParadeDB support, use the [paradedb](https://github.com/paradedb/charts/) helm chart, which as of now is an up-to-date fork of the cloudnative-pg cluster chart.

PR #384, opened 2024-09-10
Define keyNames to backups.secrets

### LINKED ISSUE #383: [backups] Allow specify the secret keys if an existing secret is created.
When you pass the following values:
```yaml
backups:
  secret:
    create: false
    name: my-existing-secret
```

Forces `my-existing-secret` to contain `ACCESS_KEY_ID` and `ACCESS_SECRET_KEY`.
The name of those keys are hardcoded in the template (See [ACCESS_KEY_ID](https://github.com/cloudnative-pg/charts/blob/f4a74a3d2873618974aacc3577f1b5c5707ab7ce/charts/cluster/templates/_barman_object_store.tpl#L28) and [ACCESS_SECRET_KEY](https://github.com/cloudnative-pg/charts/blob/f4a74a3d2873618974aacc3577f1b5c5707ab7ce/charts/cluster/templates/_barman_object_store.tpl#L31))

This could be useful in some sceneries. For instance, when you use Rook(Ceph) to create a Bucket a secret is created with the credentials, right now isn't possible to refer to that secret because the name of the keys doesn't match with the expected by this chart.

ercpereda:
Hi, this is my first contribution so please let me know if I have missed anything.

This PR adds the posibility to define the name of the keys for the secret containing the credentials of the backups.

Fixes #383


itay-grudev:
@ercpereda We require that you signoff on your commits? Can you please take a look at this: https://github.com/cloudnative-pg/charts/pull/384/checks?check_run_id=29909669764

ercpereda:
@itay-grudev Thank you for the response. The commit is now signoff.

ercpereda:
@itay-grudev Something else that can be done to get this merge? Thank you?

itay-grudev:
A local negative gravitational anomaly. Something that accelerates time, but just in my frame of reference.

jorgebsa:
This would be great to have, ran into this limitation when using an ObjectBucketClaim with storage class `openshift-storage.noobaa.io`, which prefixes the keys with `AWS_` on the generated secret. Had to resort to FluxCD's HelmRelease's `valuesFrom` to pass the values to the chart

itay-grudev:
@ercpereda Would you please be able to add a test for this in: https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

PR #387, opened 2024-09-12
Add empty string to the enum of allowed values for wal/backup encryption

brouberol:
The [documentation](https://cloudnative-pg.io/documentation/1.24/cloudnative-pg.v1/#postgresql-cnpg-io-v1-DataBackupConfiguration) as well as the CRD itself mentions that empty string is a valid value, but it was missing from the property enum, causing some CI validation error on a project I'm working on (https://integration.wikimedia.org/ci/job/helm-lint/20336/console)

```
Cluster pg-basic is invalid: For field spec.backup.barmanObjectStore.wal.encryption: spec.backup.barmanObjectStore.wal.encryption must be one of the following: "AES256", "aws:kms" - For field spec.backup.barmanObjectStore.data.encryption: spec.backup.barmanObjectStore.data.encryption must be one of the following: "AES256", "aws:kms"
```

### DIFF
--- a/charts/cloudnative-pg/templates/crds/crds.yaml
+++ b/charts/cloudnative-pg/templates/crds/crds.yaml
@@ -1689,6 +1689,7 @@ spec:
                             enum:
                             - AES256
                             - aws:kms
+                            - ''
                             type: string
                           immediateCheckpoint:
                             description: |-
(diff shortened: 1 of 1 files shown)


phisco (commented):
Firstly, CRDs here are just a copy of the one generated from the cloudnative-pg/cloudnative-pg repository, so please open an issue there. Secondly, given that it's not a required field, empty string should be already accepted, no need to add that to the enum, could be that the validating webhook is rejecting that and should be fixed.

brouberol:
Thanks for the feedback! I'll move the discussion there.

PR #389, opened 2024-09-16
Re-order template keywords

MoonMoon2:
Issue was because YAML output for this section of the cluster.yaml was

parameters:
  parameters: {}
  pg_hba: []
  pg_indent: []

Whereas it should be

parameters: {}
pg_hba: {}
pg_indent: {}

but parameters is a helm keyword so it would consume all the keywords in the parameters step before pg_hba or pg_indent were called.

This simple re-ordering should fix the issue (and has for me in testing)

Closes: #408

### DIFF
--- a/charts/cluster/templates/cluster.yaml
+++ b/charts/cluster/templates/cluster.yaml
@@ -60,12 +60,12 @@ spec:
       {{- toYaml . | nindent 6 }}
       {{- end }}
     {{- with .Values.cluster.postgresql }}
-    parameters:
-      {{- toYaml .parameters | nindent 6 }}
     pg_hba:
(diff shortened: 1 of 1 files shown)


itay-grudev (approved):

hakro:
Hi !
Any ETA for when the next release containing this fix will be published ? Just to know if I hack around the chart manually, or if I wait and then update it from upstream.

Thanx

itay-grudev:
Wait for upstream. There's a bunch of people asking me for a release so I'll try to create one soon.

P.S. Just be advised that it will contain some small breaking changes. I'll document everything in the release notes.

Issue #326, opened 2024-07-02
feat(cluster): Add support for passing inheritedMetadata to Cluster spec


paraddise:
Operator supports passing metadata that will be inherited by all resources related to Cluster.

Spec - https://github.com/cloudnative-pg/cloudnative-pg/blob/main/api/v1/cluster_types.go#L254

paraddise:
It's not stale, mentioned issues are waiting for approval

PR #400, opened 2024-09-24
feat(cluster): Add support for passing inheritedMetadata to Cluster spec

codelite7:
This should resolve https://github.com/cloudnative-pg/charts/issues/326


itay-grudev (changes requested):
Would you please be able to add a test for this in: https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

PR #412, opened 2024-10-04
Add nodeMaintenanceWindow to values and cluster template

levissfuture:


itay-grudev (changes requested):
Can you wrap it with a `with` so that it doesn't get applied if unspecified?

Issue #428, opened 2024-10-29
Prometheus rule CNPGCLusterOffline doesn't display the cluster name


mymasse:
The rule `CNPGClusterOffline` description makes use of the [`{{ .labels.job }}`](https://github.com/cloudnative-pg/charts/blob/main/charts/cluster/prometheus_rules/cluster-offline.yaml#L7) but the expression does not return this label.

So the generated alerts description becomes:

```
CloudNativePG Cluster "" has no ready instances.

Having an offline cluster means your applications will not be able to access the database, leading to
potential service disruption and/or data loss.
```

PR #429, opened 2024-10-29
fix(cluster): correct CNPGClusterOffline rule description

mymasse:
Use `namespace/cluster` instead of `labels.job` which is not returned in the expression.

Fixes: #428


mymasse:
Both `cluster-ha-critical.yaml` and `cluster-ha-warning.yaml` will work as is because `{{ .labels.job }}` does exist.

Do you want me to modify them anyways to make it consistent with the others?

itay-grudev:
@mymasse No. It's fine. Thank you very much.

itay-grudev (approved):

Issue #395, opened 2024-09-19
Inplement option to set EnablePDB


emagiz:
According to https://github.com/cloudnative-pg/cloudnative-pg/pull/3837/files#diff-67160e3b67586b511da3b6f279d73402e5bc70d9303786ddfda34aa99daae330R79, you can now set `EnablePDB` on the `Cluster` object to enable or disable PDB generation. This repository does not allow to set this yet, making it very difficult to perform maintanance. Specifically, the comments on https://github.com/cloudnative-pg/cloudnative-pg/issues/2570 are reasons to enable this in the chart.

cterence:
Up on this issue please, ArgoCD now reports PDB as a degraded resource if it has 0 allowed disruptions (https://github.com/argoproj/argo-cd/issues/20171).

PR #431, opened 2024-10-30
feat(cluster): added support for enablePDB option

mbugeia:
Add support for enablePDB configuration option

Fix https://github.com/cloudnative-pg/charts/issues/395


itay-grudev:
@mbugeia Could you add the option to the non default configuration test that verifies that options are passed correctly?

https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

You'll need to set it to `false` in the cluster values and add the option in the assertion.

mbugeia:
@itay-grudev Thanks for the review, I've added the test. Unfortunatly I'm not sure if it works as I don't know how to run the tests locally :see_no_evil:

itay-grudev (approved):

Issue #426, opened 2024-10-25
ArgoCD is forever stuck on updating the Cluster resource generated by the cluster chart


spantaleev:
I'm using the following ArgoCD application:

```yml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: cloudnative-pg-cluster-something-postgres
  namespace: argocd
  finalizers:
  - resources-finalizer.argocd.argoproj.io
spec:
  destination:
    namespace: something
    name: in-cluster
  source:
    repoURL: 'https://cloudnative-pg.io/charts/'
    targetRevision: 0.1.0
    helm:
      values: |
        fullnameOverride: cluster-postgres

        version:
          postgresql: "17"

        cluster:
          instances: 3

          storage:
            size: 3Gi

          initdb:
            database: something
            owner: something

          enableSuperuserAccess: true
          superuserSecret: cluster-postgres-superuser

          walStorage:
            enabled: true
    chart: cluster
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
  sources: []
  project: common
```

This generates a `Cluster` resource with `pg_hba: []` and `pg_ident: []`:

```yml
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  labels:
    app.kubernetes.io/instance: cloudnative-pg-cluster-something-postgres
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: cluster
    app.kubernetes.io/part-of: cloudnative-pg
    argocd.argoproj.io/instance: cloudnative-pg-cluster-something-postgres
    helm.sh/chart: cluster-0.1.0
  name: cluster-postgres
  namespace: something
spec:
  affinity:
    topologyKey: topology.kubernetes.io/zone
  bootstrap:
    initdb:
      database: something
      owner: something
      postInitApplicationSQL: null
  enableSuperuserAccess: true
  imageName: ghcr.io/cloudnative-pg/postgresql:17
  imagePullPolicy: IfNotPresent
  instances: 3
  logLevel: info
  managed: null
  monitoring:
    disableDefaultQueries: false
    enablePodMonitor: true
  postgresGID: 26
  postgresUID: 26
  postgresql:
    parameters: {}
    pg_hba: []
    pg_ident: []
    shared_preload_libraries: null
  primaryUpdateMethod: switchover
  primaryUpdateStrategy: unsupervised
  priorityClassName: null
  storage:
    size: 3Gi
    storageClass: null
[...]

PR #441, opened 2024-11-05
Fix ArgoCD diff on postgresql keys

bozorgmehr96:
Use `with` for every key in the cluster's `spec.postgresql` to prevent generating empty keys.

Fixes: #426


bozorgmehr96:
If there are other problems like this, give me the key names, and I will fix them.

pschichtel:
@bozorgmehr96 this is my ignoreDifferences rule:

```yaml
group: postgresql.cnpg.io
kind: Cluster
jsonPointers:
- /spec/bootstrap/initdb/postInitApplicationSQL
- /spec/managed
- /spec/postgresql/shared_preload_libraries
- /spec/priorityClassName
- /spec/postgresql/pg_hba
- /spec/postgresql/pg_ident
```

You covered the last two for me

bozorgmehr96:
@pschichtel Thanks. I wrapped them in if blocks.

pschichtel:
@bozorgmehr96 one more: https://github.com/cloudnative-pg/charts/blob/fb45e4ff16ca653f9d26889084f46e36cfa2d550/charts/cluster/templates/scheduled-backups.yaml#L13

bozorgmehr96:
@pschichtel, can you provide a repro? It has a non-empty default value, so it shouldn't have the problem I'm trying to fix here.

pschichtel:
@bozorgmehr96 I customized the schedule, so I don't get the defaults since it's an array. According to https://github.com/cloudnative-pg/charts/blob/fb45e4ff16ca653f9d26889084f46e36cfa2d550/charts/cloudnative-pg/templates/crds/crds.yaml#L15655 and https://github.com/cloudnative-pg/charts/blob/fb45e4ff16ca653f9d26889084f46e36cfa2d550/charts/cloudnative-pg/templates/crds/crds.yaml#L15575 the method field is not required

Stevenpc3:
I am getting a few items in argocd that are oddly just missing on the live manifest.  Not sure if it's the CRD/operator or something with values that might be false by default removed or if these values don't exist anymore?

{63806B04-DB46-4FBC-A9BF-0B554DCE9E79}

* bypassrls
* createrole
* replication
* superuser

There might be more since we didn't set them all.

itay-grudev (approved):

PR #445, opened 2024-11-11
Make container securityContext optional for OpenShift compatibility

BlueCog:
**Summary**:

This pull request adds a new configuration option to make the container-level securityContext optional in the CloudNativePG Helm chart. This change improves compatibility with OpenShift by allowing users to disable the container `securityContext` when deploying the operator.

**Background**:

When deploying the CloudNativePG operator on OpenShift, the predefined container `securityContext` can cause conflicts due to OpenShift's strict security policies. OpenShift manages user and group IDs dynamically and does not allow containers to run as a specific user or group unless explicitly permitted. Hardcoded  `runAsUser` and `runAsGroup` values (values.yaml) in the container securityContext may prevent pods from starting on OpenShift clusters.

**Changes Introduced:**

- Added a new boolean flag `enableSecurityContext` under `containerSecurityContext` in `values.yaml`, defaulting to `true` to maintain existing behavior.
- Modified `deployment.yaml` to include the container `securityContext` only when `enableSecurityContext` is set to `true`.
- Updated the `README.md` to document the new configuration option and guide users on how to use it.

**Benefits**:

- Flexibility: Users can now disable the container securityContext when deploying to OpenShift or other environments with strict security constraints.
- Compatibility: Resolves deployment issues on OpenShift by preventing conflicts with the platform's security policies.
- Maintainability: Maintains the default behavior for existing users while providing an option for those who need it.

**Usage Example**:

To disable the container `securityContext` on OpenShift, set `enableSecurityContext` to` false` in your custom values/override file:

```
containerSecurityContext:
  enableSecurityContext: false
```

**Testing**:

- Verified that the operator deploys successfully on OpenShift with enableSecurityContext set to false.
- Confirmed that the default behavior remains unchanged when the flag is set to true.

### DIFF
--- a/charts/cloudnative-pg/README.md
+++ b/charts/cloudnative-pg/README.md
@@ -35,7 +35,14 @@ CloudNativePG Operator Helm Chart
 | config.data | object | `{}` | The content of the configmap/secret, see https://cloudnative-pg.io/documentation/current/operator_conf/#available-options for all the available options. |
 | config.name | string | `"cnpg-controller-manager-config"` | The name of the configmap/secret to use. |
 | config.secret | bool | `false` | Specifies whether it should be stored in a secret, instead of a configmap. |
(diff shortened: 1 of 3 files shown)


itay-grudev:
@sxd I yield here to your opinion for review/approval.

PR #450, opened 2024-11-19
feat: Allow the user to set the operator update strategy

kaisoz:
Configuring the update strategy in Helm deployments allows users to optimize rollouts for their specific needs. For example, using strategies like `RollingUpdate` can minimize downtime by updating pods incrementally, while `Recreate` might be preferable for certain scenarios where full restarts are needed to ensure consistency.

This PR enables users to set the deployment update strategy via the `values.yaml` file, enhancing flexibility.


kaisoz:
I see that one of the pooler tests failed...  which is strange because this PR only targets the operator deployment.. could this be a flaky test?

leonardoce (approved):

itay-grudev (approved):

PR #456, opened 2024-12-03
chore: enable overriding pod monitor api version

### LINKED ISSUE #455: Enable overriding Pod Monitor API version
Enable users to override the Pod Monitor API version if they are using Azure managed prometheus which uses `azmonitoring.coreos.com/v1` instead of the default `monitoring.coreos.com/v1`.

cgpoh:
This PR enables users to override the Pod Monitor API version if they are using Azure managed prometheus which uses `azmonitoring.coreos.com/v1` instead of the default `monitoring.coreos.com/v1`.

Closes #455

### DIFF
--- a/charts/cloudnative-pg/templates/podmonitor.yaml
+++ b/charts/cloudnative-pg/templates/podmonitor.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 {{- if .Values.monitoring.podMonitorEnabled }}
-apiVersion: monitoring.coreos.com/v1
+apiVersion: {{ .Values.monitoring.podMonitorApiVersion }}
 kind: PodMonitor
 metadata:
   name: {{ include "cloudnative-pg.fullname" . }}
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -157,6 +157,8 @@ affinity: {}
 
 monitoring:
 
+  # -- The API version of the PodMonitor.
+  podMonitorApiVersion: monitoring.coreos.com/v1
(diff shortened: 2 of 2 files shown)


itay-grudev:
Even if we add this, you would only be able to see a very limited amount of metrics reported by the operator itself. The cluster _PodMonitor_ is created by the operator and there is no such API parameter there.

Can you not address this from Azure? Surely they would not have handicapped you like that.

cgpoh:
Thanks @itay-grudev , I checked [Azure document](https://learn.microsoft.com/en-us/azure/azure-monitor/containers/prometheus-metrics-scrape-crd) about their pod monitor and Azure requires api version to be azmonitoring in order for the metrics to be picked up. I’m still trying to find an equivalent Prometheus operator API  to add support for Azure monitoring in the cnpg project.

itay-grudev:
Not that I'd recommend it, but you can copy the `PodMonitors` and rewrite the `apiVersion` with a simple `jq` script.

In the mean time, I'd advise you to also raise a ticket in https://github.com/cloudnative-pg/cloudnative-pg. If they agree to do it (even without starting the work), I'll merge this.

I'd also recommend you look into managing Prometheus yourself. The [`kube-prometheus-stack`](https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack) chart is easy to install. I've never really had any issues with it. And I generally dislike the limitations of many managed services.

P.S. I've never tried it, but I also read good things about the [VictoriaMetrics](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/README.md) stack.

cgpoh:
Thanks @itay-grudev for your suggestions, I will raise a ticket in cloud native-pg.

PR #462, opened 2024-12-07
Fix graphana-dashboards link in operator chart README.md

silopolis:
GH page unavailable


itay-grudev:
Unfortunately this file is auto-generated and this solution won't work. The URL is correct for Helm purposes:

https://cloudnative-pg.github.io/grafana-dashboards/index.yaml

indeed points to the chart specs.

Issue #479, opened 2024-12-30
Pass customQueries.target_databases to monitoring configmap


dominikschidlowskidmde:
Currently only metrics, query and name are passed to the monitoring configmap.

We need to monitor multiple databases and need the target_databases feature from https://cloudnative-pg.io/documentation/current/monitoring/#structure-of-a-user-defined-metric

Todos:
- pass target_databases to the configmap

PR #480, opened 2024-12-30
feat(cluster): pass target_databases to custom-queries

dominikschidlowskidmde:


dominikschidlowskidmde:
This PR is linked to the issue https://github.com/cloudnative-pg/charts/issues/479
I pushed it to the branch dev/479 but the link didn't seem to work :(

dominikschidlowskidmde:
I now got the same problem for the "primary" key of the custom queries.
I changed it to just pass everything straight into the configmap.

Maybe the pipeline retry is going to clear the failed pipeline timeout ^^

dominikschidlowskidmde:
adjusted test assert. The yaml format differs slightly when it's generated with toYaml.

PR #488, opened 2025-01-06
Feature: support `backup.barmanObjectStore.s3Credentials.inheritFromIAMRole`

### LINKED ISSUE #487: No way to set `s3Credentials` to inheritFromIAMRole via cluster chart
When using S3 for barmanObjectStore backups, only secret credentials may be templated. An option to configure `inheritFromIAMRole` is not exposed

I would like to be able to reproduce the following via the helm chart:

```yaml
  backup:
    barmanObjectStore:
    # For backup, we S3 bucket to store data.
    # On this Blueprint, we create an S3 check the terraform output for it.
      destinationPath: s3://<your-s3-barman-bucket> #2
      s3Credentials:
        inheritFromIAMRole: true
```

jessebot:
Fixes #487 and adds the following parameters to values.yaml:

```yaml
backups:
  s3:
    inheritFromIAMRole: false

recovery:
  s3:
    inheritFromIAMRole: false
```

Please let me know if you need anything else from me!


itay-grudev:
Is this required? Have you specified a correct `serviceAccountTemplate`?

itay-grudev:
You are correct!

itay-grudev:
The same option exist for Azure called `inheritFromAzureAD` which I seem to haved added to values, but not in `_barman_object_store.tpl`. Could you add it too as part of this PR?

itay-grudev:
@jessebot I'm sorry about the confusion. For some reason I couldn't find it when running a full text search, but you're correct.

I'll try to release it next week along with several other things. Is that ok?

itay-grudev (approved):

PR #495, opened 2025-01-20
feat: add annotation support to CRD's

semnell:
Currently you cannot sync some CRDS with ArgoCD for example because of the size limit for client-side-apply.

With this change we can change just the CRD from being client-side applied and move it to server side instead of everything in the chart.

### DIFF
--- a/charts/cloudnative-pg/templates/crds/crds.yaml
+++ b/charts/cloudnative-pg/templates/crds/crds.yaml
@@ -3,6 +3,11 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    {{- if .Values.crds.annotations }}
+      {{- with .Values.crds.annotations }}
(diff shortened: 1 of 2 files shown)


charts/cloudnative-pg/templates/crds/crds.yaml
```diff
@@ -3,6 +3,11 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    {{- if .Values.crds.annotations }}
```
phisco, line 6: You can drop the if, with is already a conditional

phisco:
You need to run `make schema docs` to generate the schema and the docs. You'll have to install the necessary tools

itay-grudev:
Modifying the CRDs is a bad idea. They are not supposed to be changed.

I'd advise that you look for a solution within Argo.

You can skip the CRD installation with `crds.create=false` and manage them yourself.

PR #505, opened 2025-01-29
add imageregistry prefix to timescaledb images

gugu:
Right now timescaledb does not support container proxies, so I've added support for common .global.imageRegistry chart parameter to remove dependency on dockerhub


gugu:
@itay-grudev, feel free to contact me if you have any questions/requests for this PR

itay-grudev:
@gugu If we are adding this can you add it everywhere? Otherwise it is a bit confusing what it applies to.

Also, would you please add a test for this in: https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

PR #507, opened 2025-01-31
feat( cluster ): Added support for specifying managed services

itay-grudev:


itay-grudev:
Closed in favor of #478

PR #409, opened 2024-10-01
feat(cluster) adds configuration options for services and ldap configuration

yremmet:
This PR add the option to:
- configure service resources as described in:  https://cloudnative-pg.io/documentation/1.24/service_management/
- adding a ldap configuration

PR #509, opened 2025-01-31
feat(cluster): add ldap configuration and documentation

pha91:
This PR adds the possibility to set the LDAP Configuration which is documented [here](https://cloudnative-pg.io/documentation/1.20/postgresql_conf/#ldap-configuration)

Same as in #409 without the merged service config feature.


pha91:
Hi @itay-grudev,
we're migrating to CNPG for our project and this is a important requirement for :( 
Could you please have a look at it? 

Many thanks

itay-grudev:
I'll prioritise it. There is already one other PR for LDAP #409, but it didn't have tests unlike yours, which makes my life easier. Thank you.

itay-grudev (approved):

PR #525, opened 2025-02-24
Allow setting timeout/delay values

OmarGadd:


itay-grudev:
Make sure you add it to the non-default-configuration test to verify the properties are passed through.

OmarGadd:
@itay-grudev I have processed your feedback. Are there other to process for the PR to pass the review?

Issue #537, opened 2025-03-19
helm 3 cannot find crds


restingbull:
Helm version: `version.BuildInfo{Version:"v3.17.1", GitCommit:"980d8ac1939e39138101364400756af2bdee1da5", GitTreeState:"clean", GoVersion:"go1.23.6"}`

Chart version: `0.23.2`
```
$ helm show crds
$
```

restingbull:
Hm. Ok, this is ugly. helm 3 doesn't support templating crds.

Which means moving the crds becomes backwards incompatible with `crds.create = true`

itay-grudev:
The current behavior is beneficial for the ordinary user as it makes sure the CRDs are upgraded along with the operator through the helm chart.

If you need to manage the CRDs in another way, set the `crds.create=false` flag.

You can fetch the CRDs for a given version of the operator by running:

```bash
CNPG_VERSION=1.25.1
kustomize build https://github.com/cloudnative-pg/cloudnative-pg/config/helm/\?ref\=v$CNPG_VERSION >> crds.yaml
```

itay-grudev:
@restingbull We also don't need a solut

restingbull:
> If you need to manage the CRDs in another way, set the crds.create=false flag.

@itay-grudev 

Would you be open to a pr that adds conditionals for `Deployment`, `Service`, and `ConfigMap` to install the crds independently?

itay-grudev:
Why? What would it improve? What bug would it fix?

restingbull:
The one where the charts currently create `Deployment`, `Service`, and `ConfigMap`. Currently, if you want to use the chart as a dependency, and create a `Cluster` resource you must either pre-install the CRDS outside of helm. Otherwise, if you attempt to use the chart to install the CRDS, you end up creating the  `Deployment`, `Service`, and `ConfigMap`.

It's bit a non-ideal, and makes the chart less of a reusable building block. My goal is have a reusable stack to that configure a cluster without an undue amount of manual steps.

itay-grudev:
I'm sorry, but I'm missing the point. Why doesn't it work by just installing the chart? CRDs and everything else at the same time?

PR #539, opened 2025-03-19
Move crds directory to the place that helm 3 expects

restingbull:
A slightly ambigious wording in the helm [crd docs](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds) can lead one to expect the crds directory to exist inside `templates`.

One can, however, check to see if helm understands where the crds live by using `helm show crds`.

Before this pr:
```
$ helm show crds charts/cloudnative-pg
$
```

After:
```
$ helm show crds charts/cloudnative-pg
description: |-
{{- if .Values.crds.create }}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition ...
```

fixes: https://github.com/cloudnative-pg/charts/issues/537


restingbull:
> A slightly ambigious wording in the helm [crd docs](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds) can lead one to expect the crds directory to exist inside `templates`.
> 
> One can, however, check to see if helm understands where the crds live by using `helm show crds`.
> 
> Before this pr:
> 
> ```
> $ helm show crds charts/cloudnative-pg
> $
> ```
> 
> After:
> 
> ```
> $ helm show crds charts/cloudnative-pg
> description: |-
> {{- if .Values.crds.create }}
> apiVersion: apiextensions.k8s.io/v1
> kind: CustomResourceDefinition ...
> ```
> 
> fixes: #537

Oh dear. This turns out to be backwards incompatible change.

phisco:
helm doesn't handle in-place CRD updates, we have to keep it in the templates folder otherwise it'll fail on upgrades as soon as we add a field or change anything in the CRD. That's a known limitation of helm.

restingbull:
> helm doesn't handle in-place CRD updates, we have to keep it in the templates folder otherwise it'll fail on upgrades as soon as we add a field or change anything in the CRD. That's a known limitation of helm.

Hm. Helm is being unhelpful here -- keeping the crds in templates forces [method 2](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts), over [method 1](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you). 

Unpleasant packaging issue.

With that in mind, why keep the CRDs in the same chart as the operator?

itay-grudev:
@restingbull How do you propose distributing them otherwise? This approach allows us to make sure you have the appropriate CRD versions when updating the operator. If you need to manage them in some other way depending on your environment you can always set the `crds.create=false` flag.

itay-grudev:
Even with a subchart you still have the limitation that Helm doesn't upgrade CRDs.

PR #558, opened 2025-04-14
Cluster Chart: Add possibility to recover a specific database & owner for type object_store & backup

### LINKED ISSUE #557: Cluster Chart: Not possible to recover a specific database & owner for type object_store & backup
With the current values it is not possible to configure the properties `spec.bootstrap.recovery.database` and `spec.bootstrap.recovery.owner` of a `Cluster` when using `recovery.method: object_store` or  `recovery.method: backup` - check [here](https://github.com/cloudnative-pg/charts/blob/main/charts/cluster/templates/_bootstrap.tpl#L84-L94), while this is possible when using `recovery.method: pg_basebackup` - check [here](https://github.com/cloudnative-pg/charts/blob/main/charts/cluster/templates/_bootstrap.tpl#L34-L39). 

This makes it currently impossible to restore a certain database and owner from such backups to the `[cluster]-app` secret which instead contains credentials for the default `app` db and `app` owner as described [here](https://github.com/cloudnative-pg/cloudnative-pg/pull/2957).

In general, I would suggest that the values for `recovery` should contain a subkey to configure each `recovery.method`separately, as is already the case for the `pg_basebackup` method:

```yaml
recovery:
  method: object_store
  objectStore:
    provider: ...
    ...
  backup:
    name: ...
    ...
  pgBaseBackup:
    database: ...
    ...
```

fpfuetsch:
Closes #557 and fixes mixed up docs for `recovery.pgBaseBackup.secret/owner`

### DIFF
--- a/charts/cluster/templates/_bootstrap.tpl
+++ b/charts/cluster/templates/_bootstrap.tpl
@@ -92,6 +92,12 @@ externalClusters:
     {{- else if eq .Values.recovery.method "object_store" }}
     source: objectStoreRecoveryCluster
     {{- end }}
+    {{ with .Values.recovery.database }}
+    database: {{ . }}
+    {{- end }}
+    {{ with .Values.recovery.owner }}
+    owner: {{ . }}
+    {{- end }}
 
 externalClusters:
   - name: objectStoreRecoveryCluster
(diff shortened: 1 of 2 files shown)


itay-grudev:
@fpfuetsch Would you please add a test for this in: https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

fpfuetsch:
@itay-grudev I added two test cases

PR #562, opened 2025-04-18
fix: respect namespaceOverride for PrometheusRule generation

tamcore:
Currently, when the `cluster` chart is rendered with `namespaceOverride`, the generated `PrometheusRule` objects use `.Release.Namespace` for namespace filtering, which, chances are, does not correlate with the value of `namespaceOverride`.

As a result, the generated prometheus rules can't work, since they filter for the wrong namespace.


itay-grudev (approved):

itay-grudev:
This PR looks good, but I don't have edit access to rebase it and merge it. Could you please check the edit access checkbox? I also want to amend the test a bit to make sure it covers the change.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

tamcore:
Sorry, didn't see your reply :( "Allow edits by maintainers" is enabled now

Issue #272, opened 2024-04-25
Add support for more kinds of recovery targets


gpothier:
It is not currently possible to specify any type of `recoveryTarget` other than `targetTime` in the cluster bootstrap:

```
    {{- with .Values.recovery.pitrTarget.time }}
    recoveryTarget:
      targetTime: {{ . }}
    {{- end }}
```

What is your feeling about breaking changes*? I would suggest this:
```
    {{- with .Values.recovery.recoveryTarget }}
    recoveryTarget:
      {{- toYaml . | nindent 6 }}
    {{- end }}
```

Which means that:
- To get the same result as now, one would need to specify `recovery.recoveryTarget.targetTime` instead of `recovery.pitrTarget.time`
- It would be possible to specify other kinds of supported targets, eg. `recovery.recoveryTarget.targetImmediate`
- The nomenclature between the chart and the "native" Cluster would be the same, instead of introducing a new term `pitrTarget`

*Actually, it is actually simple to maintain backward compatibility if needed: instead of replacing the first snippet by the second one, we could keep both.

What do you think? Would you take a PR for this change?

itay-grudev:
I originally wanted to keep it simple, but I'm accepting PRs for this.

gpothier:
Sorry for taking so long (tomorrow will be exactly one year...), but I created a PR for that

PR #570, opened 2025-04-25
feat(recovery): generalize recovery target config

gpothier:
Replace `.Values.recovery.pitrTarget.time` with `.Values.recovery.recoveryTarget`, so that we can use any kind of recovery target.

This enables configuration of more advanced recovery options, such as:
- targetTime
- targetXID
- targetLSN
- targetName
- etc.

The previous behaviour is actually preserved, but deprecated, so there are no breaking changes.

Resolves cloudnative-pg/charts#272

### DIFF
--- a/charts/cluster/README.md
+++ b/charts/cluster/README.md
@@ -258,7 +258,7 @@ refer to  the [CloudNativePG Documentation](https://cloudnative-pg.io/documentat
 | recovery.pgBaseBackup.source.sslRootCertSecret.key | string | `""` |  |
 | recovery.pgBaseBackup.source.sslRootCertSecret.name | string | `""` |  |
 | recovery.pgBaseBackup.source.username | string | `""` |  |
-| recovery.pitrTarget.time | string | `""` | Time in RFC3339 format |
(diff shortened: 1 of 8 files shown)


itay-grudev:
If you have the time to write tests for it, I'll be able to move forward with it.

itay-grudev:
Take a look at this [PR](https://github.com/cloudnative-pg/charts/pull/558/files) for reference, as the test case is quite similar - since the options are exclusive there needs to be a separate test case for each option.

gpothier:
Thanks for you comments. I personally think it is easier to simply pass the whole recoveryTarget directly to the CR, instead of having two different schemas (one in the helm chart and one in the CR) to specify the same thing. Additionally, if the CR gets new recovery options, there is nothing else to do here, they will work out of the box. 
I'll write tests as requested.

gpothier:
Hi, is there a way to run tests locally? I did not find any info on that in the repo. I tried running the github workflow with Act (<https://github.com/nektos/act>), but I could not make it work (it fails when creating the Kind cluster apparently). Is it ok to commit my attempts and let Github run the tests until I get them right?

PR #588, opened 2025-05-17
Support ClusterImageCatalog, Cluster replicationSlots/volumeSnapshot recovery

nbr:
## cloudnative-pg

- Add a template to allow creating one or `ClusterImageCatalog`

## cluster

- Allow updating the `replicationSlots` field of the `Cluster`
- Enable recovery from a volume snapshot

### DIFF
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -199,6 +199,8 @@ monitoring:
     # -- Annotations that ConfigMaps can have to get configured in Grafana.
     annotations: {}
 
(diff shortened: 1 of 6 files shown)


charts/cloudnative-pg/values.yaml
```diff
@@ -199,6 +199,8 @@ monitoring:
     # -- Annotations that ConfigMaps can have to get configured in Grafana.
     annotations: {}
 
+clusterImageCatalogs: []
+
```
itay-grudev, line 203: Can you provide a commented example here?

Issue #575, opened 2025-04-29
Provide Helm Chart for `plugin-barman-cloud`


judahrand:
I'm using the existing Helm Charts to deploy both my clusters and the operator. With the deprecation of the native Barman Cloud support in 1.26.0 it would be great to see a chart provided to aid in the migration.

It could even be included as a dependency of the main operator chart perhaps. Similar to the grafana dashboard chart.

rouke-broersma:
There is an open pull request that implements this feature request, the work is largely done. The issue is not stale.

judahrand:
Aren't LLM agents great?!

Schwitzd:
I'm also interested in this topic

rouke-broersma:
> I'm also interested in this topic

https://github.com/cloudnative-pg/charts/pull/598

PR #598, opened 2025-05-28
feat(plugin-barman-cloud): add chart for plugin-barman-cloud

QuantumEnigmaa:
Towards https://github.com/cloudnative-pg/plugin-barman-cloud/issues/351

This PR replaces [the one created on the plugin-barman-cloud repo](https://github.com/cloudnative-pg/plugin-barman-cloud/pull/359).

Having an helm chart for thie plugin-barman-cloud will most likely be useful for a lot of people already using the CloudNativePG helm chart as they will be able to use the `plugin-barman-cloud` one as a subchart.

Closes: #575

### DIFF
--- a/charts/plugin-barman-cloud/test/simple-deployment/chainsaw-test.yaml
+++ b/charts/plugin-barman-cloud/test/simple-deployment/chainsaw-test.yaml
@@ -8,7 +8,7 @@ spec:
   timeouts:
     apply: 1s
     exec: 30s
-    assert: 5s
+    assert: 10s
     cleanup: 30s
   steps:
     - name: Install barman cloud plugin
--- a/charts/plugin-barman-cloud/test/simple-deployment/01-simple_deployment-assert.yaml
+++ b/charts/plugin-barman-cloud/test/simple-deployment/01-simple_deployment-assert.yaml
@@ -125,7 +125,7 @@ metadata:
   name: plugin-barman-cloud
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   name: plugin-barman-cloud
 rules:
@@ -187,12 +187,12 @@ rules:
   - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRoleBinding
 metadata:
   name: plugin-barman-cloud-binding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: Role
+  kind: ClusterRole
   name: plugin-barman-cloud
 subjects:
 - kind: ServiceAccount


QuentinBisson (approved):

itay-grudev:
@QuantumEnigmaa We want to implement a test suite for all Helm charts. Currently the Cluster chart has decent test coverage. I wouldn't want to release a new helm chart without one. If you have time would you be able to take a look and help us add one?

Our [test suite](https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test) uses [Kyverno Chainsaw](https://github.com/kyverno/chainsaw).

Feel free to contact me on [Slack](https://cloud-native.slack.com/team/U08LTPKUVFH).

charts/plugin-barman-cloud/Chart.yaml
```diff
@@ -0,0 +1,40 @@
...
+name: plugin-barman-cloud
+description: CloudNativePG plugin for barman cloud Helm Chart
+icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
+type: application
+version: "0.1.0"
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning, they should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.4.0"
```
dev-dsp, line 26: latest version seems to be 0.5.0 already?

QuantumEnigmaa: You're right, I updated it :)

charts/plugin-barman-cloud/templates/server-certificate.yaml
```diff
@@ -0,0 +1,37 @@
...
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: barman-cloud-server
+  namespace: {{ include "plugin-barman-cloud.namespace" . }}
+spec:
+  commonName: barman-cloud
+  dnsNames:
+  - barman-cloud
```
rouke-broersma, line 26: We should use the service name from the values here, once we do that the service name can be modified through values because a new certificate will be requested.

charts/plugin-barman-cloud/templates/secret.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: v1
+data:
+  SIDECAR_IMAGE: |
+    Z2hjci5pby9jbG91ZG5hdGl2ZS1wZy9wbHVnaW4tYmFybWFuLWNsb3VkLXNpZGVjYXI6dj
+    AuNC4w
+kind: Secret
+metadata:
+  name: plugin-barman-cloud-m76km67hd7
+  namespace: {{ include "plugin-barman-cloud.namespace" . }}
+type: Opaque
```
rouke-broersma, line 26: This contains the sidecar image name and version that should be injected by the controller. I think this can be an imageRef value that is templated into a configmap.

```suggestion
apiVersion: v1
data:
  SIDECAR_IMAGE: |
    "{{ .Values.sidecarImage.registry }}{{ .Values.sidecarImage.repository }}:{{ .Values.sidecarImage.tag | default .Chart.AppVersion }}"
kind: ConfigMap
metadata:
  name: plugin-barman-cloud-config
  namespace: {{ include "plugin-barman-cloud.namespace" . }}

```

charts/plugin-barman-cloud/values.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+replicaCount: 1
+
+image:
+  registry: ghcr.io
+  repository: cloudnative-pg/plugin-barman-cloud
+  pullPolicy: IfNotPresent
+  # -- Overrides the image tag whose default is the chart appVersion.
+  tag: ""
+
```
rouke-broersma, line 28:
```suggestion

sidecarImage:
  registry: ghcr.io
  repository: cloudnative-pg/plugin-barman-cloud-sidecar
  # -- Overrides the image tag whose default is the chart appVersion.
  tag: ""

```

charts/plugin-barman-cloud/templates/secret.yaml
QuantumEnigmaa: I also renamed the file to match its new kind

charts/plugin-barman-cloud/values.yaml
```diff
@@ -0,0 +1,161 @@
...
+dnsPolicy: ""
+
+# -- Update strategy for the operator.
+# ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
+# For example:
+#  type: RollingUpdate
+#  rollingUpdate:
+#    maxSurge: 25%
+#    maxUnavailable: 25%
+updateStrategy: {}
```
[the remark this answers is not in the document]
QuantumEnigmaa: In which context did you try that ? I deployed the chart in a simple kind cluster with the rollingUpdate strategy presented in the comment above in the values and it worked: the new pod becomes ready.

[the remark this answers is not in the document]
rouke-broersma: This is what I observe when triggering a rollout restart operation:

image

The readiness probe is unhealthy until the new pod gets the lease for some reason:

image

```
{"level":"info","ts":"2025-06-24T15:26:01.370038527Z","msg":"starting manager"}
{"level":"info","ts":"2025-06-24T15:26:01.37156252Z","msg":"starting server","name":"health probe","addr":"[::]:8081"}
{"level":"info","ts":"2025-06-24T15:26:01.372609226Z","msg":"attempting to acquire leader lease cloudnative-postgres/822e3f5c.cnpg.io..."}
```

Because the readiness probe of the new pod doesn't become ready until it can get the lease, there aren't sufficient ready pods and kubernetes doesn't send the shutdown signal to the old leader. Because of that the old leader doesn't drop the lease. 

This seems like a bug in plugin-barman-cloud leader election, but means that rolling update currently doesn't properly work as a default.

Do you not see this behavior??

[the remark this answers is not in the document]
rouke-broersma: I'm not deleting anything, I'm starting a rollout restart of the deployment.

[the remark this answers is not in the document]
QuantumEnigmaa: Oh ok my bad. Then yeah, using the rollout restart I get the same issue unfortunately. 

As this is a plugin-barman-cloud leader election issue, I guess we can't do much for it in this PR. I would add a comment for the `updateStrategy` field in the values mentioning this issue and modify the `strategy` field in the deployment as such: 
```yaml
{{- if and .Values.updateStrategy (eq .Values.updateStrategy "Recreate") }}
  strategy:
    {{- toYaml .Values.updateStrategy | nindent 4 }}
{{- end }}
```
Or even just hardcode the strategy field as "Recreate". 

WDYT ?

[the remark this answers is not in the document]
rouke-broersma: I think until the issue is fixed in the controller, we should hardcode the update strategy as Recreate, since rolling update simply doesn't work. When it is fixed, the chart can be changed to allow rolling update and even set rolling update as default.

Since you can confirm that you have the same issue, I will go ahead and submit the issue the the barman cloud plugin repo.

rouke-broersma:
@itay-grudev the metrics ClusterRole is for prometheus right? I would assume that prometheus is in another namespace, so a namespaced role would not work to grant authorisation to prometheus. Also these manifests are lifted from upstream: https://github.com/cloudnative-pg/plugin-barman-cloud/tree/main/config/rbac

I don't think you should modify the manifest intent in the helm chart without first modifying this upstream. Upstream uses ClusterRole for the metric endpoint.

itay-grudev:
Roles have to be explicitly assigned. If not assigned, they're unused.

I'm not aware of the Prometheus operator picking up `ClusterRoles` and assigning them to it's Pods. Let alone roles that bear no labels or annotations. 

Also, why would it require special permission? Services are accessible from all namespaces by default. 

Anyway. I think the upstream roles are a mistake. Probably left over from the original plugin template. They make no sense to me here.

rouke-broersma:
> Roles have to be explicitly assigned. If not assigned, they're unused.
It is assigned, metrics-auth-rolebinding is assigned to the plugin-barman-cloud SA so that it can perform the auth on the metrics endpoint

> I'm not aware of the Prometheus operator picking up `ClusterRoles` and assigning them to its Pods. Let alone roles that bear no labels or annotations.

It's not by default no, you would need to bind the reader role to the prometheus service account separately

> Also, why would it require special permission? Services are accessible from all namespaces by default.
Some projects like to protect the metrics endpoint with kubernetes authentication, which it looks like plugin-barman-cloud does 

> Anyway. I think the upstream roles are a mistake. Probably left over from the original plugin template. They make no sense to me here.

Fair, I do not have an opinion on whether or not it is or should be used, just observing that it exists in upstream which is why it exists here. It does look setup to support prometheus scrape authentication, the service monitor is also informed to use the prometheus service account for authentication with the endpoint: https://github.com/cloudnative-pg/plugin-barman-cloud/blob/main/config/prometheus/monitor.yaml

rouke-broersma:
The first one where I discovered it is eck operator, it uses the same mechanism as plugin barman cloud: https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s-securing-the-metrics-endpoint.html

charts/plugin-barman-cloud/test/simple-deployment/chainsaw-test.yaml
```diff
@@ -0,0 +1,31 @@
+##
+# This is a test that checks if all resources are correctly provisioned when requested.
+apiVersion: chainsaw.kyverno.io/v1alpha1
```
xlejakx, line 3: Do you have prerequisites to have kyvrrno installed? I think must be with condition

JakobStadlhuber:
Looking forward to this being merged.

Ashi0506:
Do we have any date for this?

gjrtimmer:
Does anybody know when we would have the barman-cloud-plugin as a helm chart?

fleaz:
> Does anybody know when we would have the barman-cloud-plugin as a helm chart?

Easy answer: As soon as this PR is merged. Ten people asking "are we there yet?" won't make it faster...

itay-grudev (approved):

itay-grudev:
I apologize for the insane delay. Thank you @QuantumEnigmaa.

rouke-broersma:
@itay-grudev I've been using the chart for a while now, it works fine other than the leader election but that issue is tracked upstream. Now we just need to update to 0.7.0: https://github.com/cloudnative-pg/plugin-barman-cloud/compare/v0.5.0...v0.7.0

Changes seem minimal, but some new options are added that should probably be made available in the helm chart during update. I probably have time soon to add this if no one else does.

itay-grudev:
@rouke-broersma It would be lovely if you have time for that. I'll prioritize your PR.

NigelVanHattum:
Thanks for the hard work @itay-grudev ! 

Can you perhaps also show some documentation on how to use this new chart in combination with the other charts? I'm a bit lost.

itay-grudev:
@NigelVanHattum The cluster chart doesn't support the plugin yet. I'm working on it.

ivanmagdic:
@pha91 the linked documentation does not explain on how to use new barman HELM chart. It still shows the installation using kubectl. I'm also confused on how to use it and if cert-manager is still required?

I would be more than happy to contribute with the documentation on how to use it, if we can get starting points or guidelines.

PR #510, opened 2025-02-03
cluster: add projected volume support

plesoun-stein:
I added projected volume support in In cluster specification
tested. 
my simple testing env:
namespace: aaa
clusterName: firstcnpg-cluster
values: 
```
  projectedVolumeTemplate:
    sources:
      - secret:
          name: crl
          items:
            - key: crl.pem
              path: crl/crl-01.pem
```
secret successfully mounted in `/projected/crl/crl-01.pem`

PR #637, opened 2025-08-08
cluster: add support for projected volumes

pha91:
Hi @itay-grudev, @sxd 
i've rebased the branch from PR #510
would be create to merge this as soon as possible cause we need this feature to import ca-certs to for ldaps


sxd:
Hi @pha91 

Yeah sorry I didn't saw your test, that should be ok since you're adding the assert

pha91:
Hi @sxd @itay-grudev,
waiting for more than a half year to that pr -.- 
Is there any option to speed up this? EDB Support isn't a help here as well -.-
We need this feature for a production use-case

PR #639, opened 2025-08-08
feat(cloudnative-pg): Setup optional PDB

jcpunk:
This permits putting a PDB around the operator deployment.


WoahitsJoe:
Thanks for opening this. We need configurable operator PDB support and found this PR while checking for existing work before contributing. Are you planning to refresh it against current `main`?

If not, I am happy to prepare a clean replacement after maintainer confirmation so we avoid duplicating or competing with your work.

jcpunk:
Feel free to close this out and go with what you think makes sense

WoahitsJoe:
Thanks for the handoff, @jcpunk. I opened #998 as a clean replacement on current main, with the operator PDB implementation, generated schema/docs, and validation coverage.

PR #657, opened 2025-08-28
feat(cloudnative-pg): add annotations to all CRDs

### LINKED ISSUE #656: CRD annotations through values.yaml
As discussed in https://github.com/cloudnative-pg/charts/issues/325
the CRDs are too big to be installed through ArgoCD without requiring ServerSideApply.

To facilitate this please create a dictionary `crds.annotations` that lets me set an annotation
`argocd.argoproj.io/sync-options: ServerSideApply=true`
exclusively for the big CRD.

Thanks!
-xmj

xmj:
We want to be able to pass annotations to all CustomResourceDefinitions.

This is helpful in that it allows passing an annotation to ArgoCD to enable ServerSideApply:

```
crds:
  annotations:
    argocd.argoproj.io/sync-options: ServerSideApply=true
```

Fixes #656


kastl-ars:
What is missing for this to be merged?

kastl-ars:
> What is missing for this to be merged?

Sorry for bringing this up again, but why is this simple PR still open? What is missing?

0xayf:
Hi.

+1 for this feature.

In my case, I use a [git generator](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Git/) to point to a repository containing a bunch of helm charts, one of them being cloudnative-pg.

Currently, without this PR I would have to set `ServerSideApply=true` on my single ApplicationSet, which then applies it all of those charts, causing issues with several of them.

In an ideal world, I'd add the annotations as @xmj has outlined; this is exactly how it is handled in other charts.

Thanks.

PR #665, opened 2025-09-21
feat(pooler): create Secret and use pooler address in JDBC URL

shusaan:
## What the Issue Is

Currently, when enabling the pooler from the Helm chart (`values.yaml`), the pooler is created successfully, but the generated Secret is **not updated** to use the pooler endpoint.

Instead, the Secret continues to expose the default PostgreSQL connection URL (`postgresql://...`) from the cluster.

This means applications cannot automatically connect via the pooler, defeating the purpose of enabling it through the chart.

This behavior has been raised in related issues:

- [cloudnative-pg/cloudnative-pg#3201](https://github.com/cloudnative-pg/cloudnative-pg/issues/3201)  
- [cloudnative-pg/cloudnative-pg#3111](https://github.com/cloudnative-pg/cloudnative-pg/issues/3111)

---

## How the Solution Works

Since handling this directly in the operator is more complex and requires deeper architectural changes, this PR provides a Helm chart level solution.

When `pooler.enabled = true` in `values.yaml`:

- A new **Secret** is generated for the pooler containing:
  - JDBC URL pointing to the pooler service  
  - username  
  - password  

- The Secret is mapped to the cluster, ensuring workloads can connect using the pooler endpoint instead of the PostgreSQL endpoint.

This approach provides a simpler and more flexible way to expose pooler connectivity without waiting for operator level changes.

---

## Current Scope

- Implemented Secret generation for a **single pooler** (e.g., `rw` pooler).  
- The generated Secret is mapped to the cluster for easy consumption by applications.  

---

## Future Improvements (possible next steps)

- Support for multiple poolers (e.g., `ro` pooler or custom poolers).  
- Configurable Secret naming convention.

### DIFF
--- a/charts/cluster/templates/_bootstrap.tpl
+++ b/charts/cluster/templates/_bootstrap.tpl
@@ -10,6 +10,13 @@ bootstrap:
     {{- if .Values.cluster.initdb.owner }}
     owner: {{ tpl .Values.cluster.initdb.owner . }}
     {{- end }}
+    {{- if .Values.poolers }}
+    secret:
+      name: {{ include "cluster.fullname" . }}-pooler-{{- range .Values.poolers }}{{- if eq .name "rw" }}{{ .name }}{{- end }}{{- end }}
+    {{- else if .Values.cluster.initdb.secret }}
+    secret:
+      {{- toYaml .Values.cluster.initdb.secret | nindent 6 }}
+    {{- end }}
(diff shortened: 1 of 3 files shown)


ludat:
IMHO this should be directly in the controller instead of the helm chart

I started implementing that a couple of months ago but I haven't finished it yet

PR #669, opened 2025-09-27
ISSUE-668 templated clusterImageCatalogs

### LINKED ISSUE #668: [Feature] Template ClusterImageCatalogs as specified in values yaml
As an operator is a cluster-wide API, and a ClusterImageCatalog is also a cluster-wide API, it makes sense that a user interested in a ClusterImageCatalog would prefer to install it alongside the cnpg operator in a single helm install.

This way the user can install the cnpg operator, and then immediately reference any needed ClusterImageCatalog without having to install it standalone in a separate helm install, or even worse as part of a helm install for a namespaced cluster.

blaisemGH:
## Summary

Adds support for provisioning ClusterImageCatalog resources alongside the operator via a new clusterImageCatalogs values key.

Since the operator is cluster-scoped and ClusterImageCatalog is also cluster-scoped, it makes sense to install them together in a single helm install, rather than requiring a separate kubectl apply step.

Closes #668

## Changes
- New template ClusterImageCatalog.yaml with license header, standard labels, commonAnnotations support
- Input validation via Helm required function for clear error messages
- JSON schema validation in values.schema.json
- Default clusterImageCatalogs: [] with commented example in values.yaml

## Prior art
Aware of #588 which implements the same feature. This PR builds on that approach with additional improvements:
- Apache 2.0 license header (matching all other templates)
- required validation for clear template-time error messages
- | quote on image values for YAML safety
- Strict integer type for major in the JSON schema (matching the CRD spec)

### DIFF
--- a/charts/cloudnative-pg/templates/ClusterImageCatalog.yaml
+++ b/charts/cloudnative-pg/templates/ClusterImageCatalog.yaml
@@ -1,0 +1,36 @@
+#
+# Copyright The CloudNativePG Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


timvtinsa:
Any chance this can be integrated ? Or is there any plan to offer the possibility to deploy a **ClusterImageCatalog** using the **cluster** helm chart @itay-grudev ?

PR #671, opened 2025-09-30
Add replica mode with pg_basebackup and WAL repl.

antevens:
Adds the ability to create clusters in replica mode. Refactors existing branch https://github.com/cloudnative-pg/charts/tree/feat/replica-cluster to use externalCluster template and add pg_basebackup mode.


antevens:
Hi @itay-grudev, 

I was trying to use that branch myself and found a couple of issues, I might just have missed something.

1. It doesn't merge cleanly to main
2. It only supports WAL based replicas, no pg_basebackup or streaming

My first iteration was to add pg_basebackup based replicas and I'm working on adding streaming replicas based on the following documentation.

https://cloudnative-pg.io/documentation/1.27/replica_cluster/

Let me know if I misunderstood or missed something.

itay-grudev:
It does support pg_basebackup replicas. See: https://github.com/cloudnative-pg/charts/blob/07f8c32800f2d66ad230ef078ce01b4561b32301/charts/cluster/values.yaml#L514-L531

I'll update the PR and resolve its conflict.

itay-grudev:
As well as this one that does a hybrid configuration: https://github.com/cloudnative-pg/charts/blob/feat/replica-cluster/charts/cluster/test/replica/08-replica_hybrid_cluster.yaml

itay-grudev:
@antevens That being said, if I am missing something, feel free to correct me.

antevens:
Thanks, I missed the tests and probaly mis-understood your code and the docs.
I'll re-test with your branch tomorrow but based on a quick test it looks like the rendered cluster resource is exactly as it should be.

PR #686, opened 2025-10-08
add minutes, seconds and remove isCA

zoe-tek:
Certificate resource removes isCA: false as it is the default and adds "0m0s" to the duration and renewBefore.
 This change will allow a clean sync in argoCD.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,4 @@
+# Security Policy
+
+For the most current and detailed version of our security procedures, please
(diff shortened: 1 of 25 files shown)


zied-jt (approved):
LGFM
Source: https://cert-manager.io/docs/usage/certificate/ and https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateSpec
> Value must be in units accepted by Go time.ParseDuration https://golang.org/pkg/time/#ParseDuration.

PR #703, opened 2025-10-27
feat(plugin-barman-cloud): add additional annotations to service account

QuantumEnigmaa:
This PR adds the possibility for a user to add additional annotations to the service account if it's created from the chart.

This is especially useful when using [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) authentication on AWS kubernetes clusters as the user can add the annotation directly in the chart's values.


QuentinBisson:
Hey @itay-grudev do you think you could take a look on your side?

itay-grudev (approved):

Issue #652, opened 2025-08-21
Misalignment Between Helm values schema and CRD for pg_basebackup.secret


rogiervandergeer:
The Helm chart currently expects `.Values.cluster.recovery.pgBaseBackup.secret` to be a string, but when rendered into the Cluster resource for CloudNativePG, this field must be an object conforming to the `LocalObjectReference` type — i.e., it should be rendered as:

```yaml
secret:
  name: my-secret
```

However, the chart renders:
```yaml
secret: my-secret
```
This causes a validation failure from the CNPG admission webhook:

```
json: cannot unmarshal string into Go struct field BootstrapPgBaseBackup.spec.bootstrap.pg_basebackup.secret of type api.LocalObjectReference
```

0xfabio:
I have tried to fix this in a PR. Any chance we could get a review here?

PR #736, opened 2025-11-24
fix( cluster ): recovery.pgBaseBackup.secret not applied correctly

itay-grudev:
This fixes the `recovery.pgBaseBackup.secret` parameter by correctly setting it as a `LocalObjectReference`.

I also decided to rename the parameter to: `recovery.pgBaseBackup.secretName` to better reflect that this is just the secret name.

The non-backwards compatible name change is acceptable here, because the parameter never worked.

Closes: #652 #698

### DIFF
--- a/charts/cluster/values.yaml
+++ b/charts/cluster/values.yaml
@@ -474,6 +474,36 @@ backups:
   # -- Retention policy for backups
   retentionPolicy: "30d"
 
+##
+# Database management configuration
+databases: []
+ # - name: app                      # -- Name of the database to be created.
+ #   ensure: present                # -- Ensure the PostgreSQL database is present or absent - defaults to "present".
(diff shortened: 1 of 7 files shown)


charts/cluster/values.yaml
```diff
@@ -103,8 +103,8 @@ recovery:
   pgBaseBackup:
     # -- Name of the database used by the application. Default: `app`.
     database: app
-    # -- Name of the secret containing the initial credentials for the owner of the user database. If empty a new secret will be created from scratch
-    secret: ""
+    # -- Name of the secret containing the initial credentials for the owner of the user database. If empty a new secret will be created from scratch. The secret should include
```
phisco: "The secret should include"... what?

PR #737, opened 2025-11-24
feat(plugin-barman-cloud): make actual use of issuer name from values

### LINKED ISSUE #702: Variable `.Values.certificates.issuerName` not used
The values file for `plugin-barman-cloud` has the variable `certificates.issuerName` but it's not used for generating the name of the issuer, which instead simply uses `{{ include "plugin-barman-cloud.fullname" . }}-selfsigned-issuer`.

### LINKED ISSUE #893: plugin-barman-cloud helm chart exposes value to set the name of the cert issuer, but doesnt actually use it when rendering the certificate resources.
I was trying to configure the helm chart to use my own cluster issuer instead of the self signed one. I saw in default values it exposes "issuerName" in the certificates block but currently its not used when templating the cert resources. It would be nice to make it possible to configure a custom issuer/clusterIssuer to use for the client/server certs.

pascaliske:
Fixes #702, Fixes #893.

### DIFF
--- a/charts/plugin-barman-cloud/templates/_helpers.tpl
+++ b/charts/plugin-barman-cloud/templates/_helpers.tpl
@@ -71,3 +71,10 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Create the name of the certificate issuer to use
+*/}}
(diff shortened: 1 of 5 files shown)


pascaliske:
@gbartolini I updated the branch based on the current main - is there something else I can do to get this thing merged? 🙂

Issue #673, opened 2025-10-03
plugin-barman-cloud: Allow omitting `containerSecurityContext.runAs*`


ThisIsntTheWay:
At present (v0.2.0), the chart forces us to set `containerSecurityContext.runAs*` as they are `required` according to the schema: https://github.com/cloudnative-pg/charts/blob/05cd9a5288427e589362c59ee0d7520468c16cb4/charts/plugin-barman-cloud/values.schema.json#L148-L155

Unfortunately, this creates issues for deployments on OpenShift where a random UID and GID in a given range is assigned to pods and setting `runAs*` to values other than that range requires the use of SCCs.  
As such, we are using kustomize patches to get rid of said fields in the rendered deployment.

It would be much more convenient if we could pass something like this as values:
```yaml
containerSecurityContext:
  runAsUser: null
  runAsGroup: null
```

ThisIsntTheWay:
Chart version `0.7.0` has now incorporated #742, allowing us to set the following in the chart values:
```yaml
containerSecurityContext:
  runAsUser: null
  runAsGroup: null
```

:shipit: 👍

PR #740, opened 2025-11-27
feat(plugin-barman-cloud): add possiblity to turn off securityContext on pod- and container level

Pidu2:
This PR addresses the requirements of https://github.com/cloudnative-pg/charts/issues/673.

In an OpenShift environment, you usually do not want to set `runAsUser` or `runAsGroup` to something specific, instead it is chosen automatically by the cluster to a random value. Also, any other SecurityContext Values of the Pods will automatically be set to secure defaults (`allowPrivilegeEscalation: false` etc.).

It is therefore usually best to just not set it at all.

With this change, `containerSecurityContext.enabled` and `podSecurityContext.enabled` can be set to `false` to stop rendering out these fields completely.

In https://github.com/cloudnative-pg/charts/issues/673 it was proposed to just remove the `required` setting for `runAsUser` or `runAsGroup`. If this is considered preferred to my solution, I can adjust the PR accordingly.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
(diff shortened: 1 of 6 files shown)


Pidu2:
You're right @Reamer , readOnlyRootFilesystem is not set. I will open another PR.
I wanted to go this way because the charts seems to ensure that everything that is set in default `values.yaml` is also contained in the schema...

PR #749, opened 2025-12-11
Barman Plugin: Conditionally create self-signed Issuer based on certificate settings

rezaxd:
Update on Helm template for the Issuer resource so it is created only when certificate generation is explicitly enabled. This caused unnecessary issuer creation even when no client or server certificates were required.


itay-grudev:
Resolved in #811

PR #755, opened 2025-12-16
fix(metrics): make port configurable

chomatdam:
This change makes the Prometheus metrics port configurable in the cloudnative-pg Helm chart (instead of being hardcoded to 8080).


chomatdam:
let me know if there is anything I can do to help this PR getting merged, otherwise I'll just wait your backlog gets to it

geoffo-dev:
I was just about to add this change too..! Is there any chance this could be merged?

PR #758, opened 2025-12-18
Backup and recovery region in values.yaml not being used as expected

ChrisTomAlxHitachi:
In my testing I noticed the `region` value in `Values.yaml` seen in the following places are defaulting to `us-east-1`
- `backups.s3.region`
- `recovery.s3.region`

When testing in region `us-east-1` I didn't face issues with either backup or recovery but when testing with region `boston` I noticed backups failing.

On postgres I saw errors like
```
{"level":"info","ts":"2025-12-03T22:04:36.428619311Z","logger":"barman-cloud-check-wal-archive","msg":"2025-12-03 22:04:36,428 [1447] ERROR: Barman cloud WAL archive check exception: An error occurred (400) when calling the HeadBucket operation: Bad Request","pipe":"stderr","logging_pod":"postgres-cluster-1"}
```

On my custom s3 I saw errors like
```
AUTHORIZATION_REGION_MISMATCH_ERROR
```


itay-grudev:
This does not seem like the right solution for your issue. The env var `REGION` should not have any effect. The AWS CLI specifically looks for: `AWS_DEFAULT_REGION` or `AWS_REGION`. Your issue was likely resolved by something else.

PR #773, opened 2026-01-02
test(cluster): Add support for extension upgrade test under cluster/test/database-management

philippemnoel:
Over at @paradedb we have been using the CNPG `cluster` Chart as a baseline for our chart for quite some time. We had some tests around upgrading database extensions in `test/database-management`.

In an effort to 1) contribute our changes to the community and 2) reduce our diff with upstream, I'd like to propose adding these tests to the upstream charts. They confirm that extensions upgrade properly via the database management functionality, and ensure they don't regress.


charts/cluster/test/database-management/02-extension-upgrade-init-assert.yaml
```diff
@@ -0,0 +1,25 @@
...
+status:
+  readyInstances: 2
+  phase: Cluster in healthy state
+---
+apiVersion: postgresql.cnpg.io/v1
+kind: Database
+metadata:
+  name: extension-upgrade-cluster-cluster
+spec:
+  name: test-db-icu
```
philippemnoel: @itay-grudev let me know if this is not right. I based it on 01-database-parameters.

PR #707, opened 2025-11-03
add livenessProbe to plugin-barman-cloud chart

nkgilley:
fixes https://github.com/cloudnative-pg/charts/issues/687

PR #776, opened 2026-01-06
fix: plugin-barman-cloud port for readiness probe

### LINKED ISSUE #710: [plugin-barman-cloud] Readiness probe does not allow HA
Hi, 

We are using the plugin-barman-cloud chart to use object storage for backups. For HA requirements, we increased the replicaCount to 2 or 3, and we notice that inactive pods do not open port 9090. They open it when they achieve lease acquisition, but until then they are marked as down, which triggers false positive alerts and noise.

Is there anything we can do to mitigate this ? 

Regards,

# Code :

https://github.com/cloudnative-pg/charts/blob/da9471f83fecd4d6aa301ad0cc34858721476df3/charts/plugin-barman-cloud/templates/deployment.yaml#L76

tuunit:
## Description

At the moment the port `9090` is used for the readiness probe which is the application port which only get opened / used by the leader. For an HA setup with multiple pods this fails because none leaders only open the health/metrics port `8081`.

Fixes #710

### DIFF
--- a/charts/plugin-barman-cloud/Chart.yaml
+++ b/charts/plugin-barman-cloud/Chart.yaml
@@ -22,7 +22,7 @@ description: Helm Chart for CloudNativePG's CNPG-I backup plugin using Barman Cl
 kubeVersion: ">=1.29.0-0"
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: "0.7.0"
+version: "0.7.1"
 appVersion: "v0.13.0"
 sources:
   - https://github.com/cloudnative-pg/plugin-barman-cloud
--- a/charts/plugin-barman-cloud/Chart.yaml
+++ b/charts/plugin-barman-cloud/Chart.yaml
@@ -22,7 +22,7 @@ description: Helm Chart for CloudNativePG's CNPG-I backup plugin using Barman Cl
 kubeVersion: ">=1.29.0-0"
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: "0.7.1"
+version: "0.7.0"
 appVersion: "v0.13.0"
 sources:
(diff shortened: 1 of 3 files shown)


dixneuf19:
Hi, also looking for this fix 🙏

tuunit:
Friendly ping @itay-grudev @mnencia

itay-grudev (approved):

itay-grudev:
If I am not mistaken, the liveness probe proposal in #707 would suffer from the same issue. Could you add it in this PR with the corrected port?

tuunit:
@itay-grudev thanks for the response! I rebased the commit from #707 into this PR and changed its port too. this way @nkgilley still gets proper attribution for his change

itay-grudev:
That's very thoughtful. Thank you very much.

itay-grudev:
For some reason I'm not able to directly edit the test in your branch, but the following part of the test needs to be updated:
https://github.com/tuunit/cloudnative-pg-charts/blob/f612cb554a1b8365dd2f584693eb92a7bf6b32c3/charts/plugin-barman-cloud/test/simple-deployment/01-simple_deployment-assert.yaml#L26

tuunit:
> For some reason I'm not able to directly edit the test in your branch, but the following part of the test needs to be updated

Done!

erikgb:
@itay-grudev, for some reason, this change snuck into release 0.7.0, which cannot be intentional. We template all our Helm charts, which allowed us to discover this. You can also see that the OCI tag was moved yesterday, ref. https://github.com/cloudnative-pg/charts/pkgs/container/charts%2Fplugin-barman-cloud. Please review your CI setup! 🙏

tuunit:
@itay-grudev indeed the release pipeline runs for every push to main:
https://github.com/cloudnative-pg/charts/actions/runs/28109064990/job/83231165632

itay-grudev:
There is indeed a bug in the OCI push action, that still hasn't been fixed. I'll see what I can do to mitigate it on our side.

The regular HTTPS repo isn't affected. Although those aren't signed.

PR #780, opened 2026-01-18
Update all the cluster chart documentation links

jessebot:
Hoi team!

All doc links were all pointing at links that gave a 404 because the URL had `documentation` instead of `docs`, and for the version it had `current` instead of an actual version. If you'd like, I can update the now hardcoded `1.28` version to `devel` so it's always the latest.

I can also bump the `version` in Chart.yaml if you need me to.

Thanks!

### DIFF
--- a/charts/cloudnative-pg/Chart.lock
+++ b/charts/cloudnative-pg/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: cluster
   repository: https://cloudnative-pg.github.io/grafana-dashboards
-  version: 0.0.2
-digest: sha256:fcf16ad357c17be3dd79c138723e78e9e101fecc5d07d9371299c32b9f85dbd9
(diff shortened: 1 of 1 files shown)


charts/cluster/values.yaml
```diff
@@ -396,11 +396,11 @@ cluster:
   serviceAccountTemplate: {}
 
   # -- Configure the Pod Security Context.
-  # See: https://cloudnative-pg.io/documentation/preview/security/
+  # See: https://cloudnative-pg.io/docs/preview/security/
   podSecurityContext: {}
 
   # -- Configure Container Security Context.
-  # See: https://cloudnative-pg.io/documentation/preview/security/
+  # See: https://cloudnative-pg.io/docs/preview/security/
```
jessebot:
```suggestion
  # See: https://cloudnative-pg.io/docs/1.28/security/
```

PR #790, opened 2026-01-21
make immediate configurable

### LINKED ISSUE #789: scheduled-backups hardcoded imediate spec
The ScheduledBackup template currently does not allow overriding the `immediate` flag.

```yaml
{{ if .Values.backups.enabled }}
{{ $context := . -}}
{{ range .Values.backups.scheduledBackups -}}
---
apiVersion: postgresql.cnpg.io/v1
kind: ScheduledBackup
metadata:
  name: {{ include "cluster.fullname" $context  }}-{{ .name }}
  namespace: {{ include "cluster.namespace" $ }}
  labels: {{ include "cluster.labels" $context | nindent 4 }}
spec:
  immediate: true
  schedule: {{ .schedule | quote }}
  method: {{ .method }}
  backupOwnerReference: {{ .backupOwnerReference }}
  cluster:
    name: {{ include "cluster.fullname" $context }}
{{ end -}}
{{ end }}
```

dejanu:
Fix #789


phisco:
@dejanu, sorry for the delayed response, however this requires a few changes:

1. The `default` filter doesn't actually let users disable `immediate` — Sprig's `default` returns the default for any zero value, and `false` is the zero value for bool. Rendering with `immediate: false` still produces `immediate: true`. So, you'd need something like, `{{ if hasKey . "immediate" }}{{ .immediate }}{{ else }}true{{ end }}`, I think.

2. The new field should also be declared in `values.schema.json` so it's documented and picked up by tooling.

itay-grudev:
In addition to the note above, I'd like to add that it should also be added to the values.yaml and the scheduled backups tests.

PR #811, opened 2026-02-17
allow certificate isssuer to be optional, wire issuerName

### LINKED ISSUE #702: Variable `.Values.certificates.issuerName` not used

### LINKED ISSUE #810: Wire certificate.issuerName and remove cert-manager dependency
At the moment, the helm chart value certificate.issuerName is not wired to any option. In addition the chart is dependent on cert-manager as the Issuer is not optional. This seems like a bug/mistake.

I suggest wiring issuerName to the template "certificate-issuer" and certificates issuer ref.

Similarly allow Cert-manager and Issuer to be optional by adding a toggle like done for clientCertificate/serverCertificate

.Values.certificate.createIssuer: true

North14:
This pull request makes the following changes to the chart for plugin-barman-cloud:

- Allows creation of issuer to be optional by adding (.Values.certificate.createIssuer) a boolean switch (and therefore removes cert-manager as a dependency)

- .Values.certificate.issuerName is previously unusued. This change wires it to the name of certificate issuer

Closes #810
Resolves #737, #749
Fixes #702, #893


itay-grudev:
@pascaliske, @North14's solution allows using a custom issuer, but your solution in #737 makes better use of issuer name, while here it's only applied as a suffix. 

I'm combining your PR's here and will merge it if passes the tests.

itay-grudev (approved):

PR #818, opened 2026-02-24
feat(cluster): Adding topologySpreadConstraints support to the cluster

### LINKED ISSUE #817: Topology spread constraints
I'd like to be able to specify the [topologySpreadConstraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) for the cluster in addition to the affinity.

miquelpuigmena:
This adds support for specifying
[`topologySpreadConstraints`](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) for the cloudnative-pg cluster.

Resolves [cloudnative-pg#817](https://github.com/cloudnative-pg/charts/issues/817)

### DIFF
--- a/charts/cluster/README.md
+++ b/charts/cluster/README.md
@@ -151,6 +151,7 @@ refer to  the [CloudNativePG Documentation](https://cloudnative-pg.io/documentat
 | backups.wal.maxParallel | int | `1` | Number of WAL files to be archived or restored in parallel. |
 | cluster.additionalLabels | object | `{}` |  |
 | cluster.affinity | object | `{"topologyKey":"topology.kubernetes.io/zone"}` | Affinity/Anti-affinity rules for Pods. See: https://cloudnative-pg.io/documentation/current/cloudnative-pg.v1/#postgresql-cnpg-io-v1-AffinityConfiguration |
+| cluster.topologySpreadConstraints | list | `[]` | TopologySpreadConstraints rules for the cluster. See: https://cloudnative-pg.io/documentation/current/cloudnative-pg.v1/#ClusterSpec |
 | cluster.annotations | object | `{}` |  |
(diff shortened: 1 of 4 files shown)


phoermandinger:
is there a time plan when this feature gets merged?

miquel-cala:
@itay-grudev could we try to have a look at this feature? It seems fairly small and relevant to the chart

ehbello:
Heads-up on a detail worth folding in before this lands, whichever shape it takes.

Several people (including #949) reach for the obvious selector when writing these constraints by hand:

```yaml
labelSelector:
  matchLabels:
    cnpg.io/cluster: <cluster-name>
```

That label is not unique to instance pods. The operator also sets it on:

- pooler (pgbouncer) pods — `pkg/specs/pgbouncer/deployments.go` (they additionally get `cnpg.io/podRole: pooler`)
- Job pods — `pkg/specs/jobs.go`, and these carry **no** `cnpg.io/podRole` at all, only `cnpg.io/jobRole`

So a constraint using that selector counts those pods towards the per-domain skew. The Job case applies to every cluster, not just ones running poolers: `initdb`, `join` and `full-recovery` all produce Job pods. The `join` Job is the awkward one — it runs precisely while a new instance is being placed, so under `whenUnsatisfiable: DoNotSchedule` with a small `maxSkew` it can count against the very instance it is preparing and leave it unschedulable.

Adding `cnpg.io/podRole: instance` (set on instance pods only, `pkg/specs/pods.go`) restricts the selector to exactly the pods the constraint is meant to spread:

```yaml
labelSelector:
  matchLabels:
    cnpg.io/cluster: <cluster-name>
    cnpg.io/podRole: instance
```

This PR passes `topologySpreadConstraints` through verbatim, so it is not strictly a bug here — users supply their own selector. But since the docs example is what people copy, it seems worth either using the two-label selector in the example, or defaulting `labelSelector` to it when the user omits it, so the safe form is the default rather than something you have to know about.

I have the defaulting variant working on a fork if it is useful: <https://github.com/maarlab-rethinking/cloudnative-pg-charts/tree/feat/cluster-topology-spread-constraints>. Happy to help rebase this PR too — it is currently showing conflicts against `main`.

PR #833, opened 2026-03-09
chore(cloudnative-pg): Enable Namespaced-Based Deployment for CNPG Operator

Charbel-Metrot:
We attempted to deploy the CNPG Operator in namespaced mode rather than as a cluster-wide deployment. During testing, we discovered that the value provided to the WATCH_NAMESPACE flag was not being picked up by the operator.

This change ensures that the CNPG Operator correctly reads the WATCH_NAMESPACE configuration so it can monitor the intended namespaces. Multiple namespaces can be specified by listing them as a comma-separated value.

Example configuration:

cloudnative-pg:
  enabled: true
  config:
    clusterWide: false
    namespaced: true
    data:
      WATCH_NAMESPACE: test

For multiple namespaces:
WATCH_NAMESPACE: namespace-1,namespace-2,namespace-3


itay-grudev:
I'm closing this as I don't see the point and it looks like an AI submission.

Charbel-Metrot:
@itay-grudev that's not an AI change but your current CNPG doesn't support namespace based deployment - it's a cluster based deployment which creates many issues in restricted environments.
That's why I pushed this suggestion to allow the CNPG operator to be namespaced when needed

charts/cloudnative-pg/templates/deployment.yaml
```diff
@@ -91,7 +91,7 @@ spec:
           value: "{{ .Values.monitoringQueriesConfigMap.name }}"
         {{- if not .Values.config.clusterWide }}
         - name: WATCH_NAMESPACE
-          value: "{{ include "cloudnative-pg.namespace" . }}"
+          value: "{{ index .Values.config.data "WATCH_NAMESPACE" | default (include "cloudnative-pg.namespace" .) }}"
```
itay-grudev, line 94: `index` is the wrong function get get a key from a dict.

itay-grudev (changes requested):

PR #841, opened 2026-03-17
fix(monitoring): guard against nil pointer in pooler PodMonitor template

philippemnoel:
## Summary
- Fixes nil pointer panic in `podmonitor-pooler.yaml` when a pooler defines `monitoring.enabled: true` without a `podMonitor` sub-key
- Go templates evaluate all arguments to `and` before calling the function, so `.monitoring.podMonitor.enabled` is evaluated even when `.monitoring.podMonitor` is nil, causing a nil pointer panic
- Replaces the single `and` check with nested `if` blocks so `.monitoring.podMonitor.enabled` is only evaluated when `.monitoring.podMonitor` is confirmed non-nil

## How to reproduce
Define a pooler with monitoring enabled but without a `podMonitor` key:
```yaml
poolers:
  - name: rw
    type: rw
    instances: 3
    poolMode: transaction
    parameters:
      max_client_conn: "1000"
      default_pool_size: "25"
    monitoring:
      enabled: true
```

This produces:
```
Error: template: cluster/templates/podmonitor-pooler.yaml:3:54: executing "cluster/templates/podmonitor-pooler.yaml" at <.monitoring.podMonitor.enabled>: nil pointer evaluating interface {}.enabled
```


charts/cluster/test/monitoring/01-monitoring_cluster-assert.yaml
```diff
@@ -17,6 +17,8 @@ spec:
         - name: monitoring-cluster-monitoring
           key: custom-queries
       enablePodMonitor: false
+status:
+  readyInstances: 2
```
itay-grudev: This test doesn't need to wait for the two instances to become ready. We are only interested in whether the associated resources like `PodMonitors`, `ConfigMaps` and `PrometheusRules` are correctly provisioned when requested. We haven't changed anything in the instance configuration so waiting for the instances is just wasting energy. The moment the other resources are available the test should be considered successful.

itay-grudev (approved):

PR #847, opened 2026-04-04
feat(cluster): expose backups.target option in helm values.

cloudymax:
Closes #846

Exposes the `backup.target` option in the helm `values.yaml` file. 
This will allow users to specify if they would like on-demand-backups to pull from the primary or a standby replica.

Supports 2 options:
- `primary`
- `prefer-standby` (default)


itay-grudev:
Yes. I like this. If you could resolve the conflicts and add the parameter in the non default configuration test I'll merge it.
https://github.com/cloudnative-pg/charts/tree/main/charts/cluster/test/postgresql-cluster-configuration

cloudymax:
@itay-grudev cleaned up and rebased on the latest main. I'm a bit confused as to where to add tot he tests as the `postgresql-cluster-configuration` manifests all have backups disabled so this option wouldn't be relevant there. I see that in the `scheduledbackups` dir that there are backups being tested :thinking:

danishedb:
Hi @cloudymax 👋
To cover the new parameter in the non-default configuration test, you can update the two files under charts/cluster/test/postgresql-cluster-configuration/:
01-non_default_configuration_cluster.yaml  and 01-non_default_configuration_cluster-assert.yaml
and you have to enable backup and pass other parameters as needed..  
```
backups:
  enabled: true
 ```

**Comment by danishedb** [comment]
I have added the parameters in this [pr](https://github.com/cloudymax/cnpg-charts/pull/1) based on this branch and tested it on my fork as well. @cloudymax Have a look.

File: charts/cluster/Chart.yaml [hunk 1]
```diff
@@ -22,7 +22,7 @@ description: Deploys and manages a CloudNativePG cluster and its associated reso
 kubeVersion: ">=1.29.0-0"
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: 0.8.1
+version: 0.8.2
```
danishedb, line 25: I think the version shall be 0.8.1 till we are releasing the chart.

PR #851, opened 2026-04-13
Add support for ImageVolume extensions

dvarrazzo:
The Cluster resource defines the `.spec.postgresql.extensions` list of objects to import an extension as an [image volume](https://cloudnative-pg.io/docs/1.29/imagevolume_extensions
).

However the chart doesn't recognise this setting and throws an error if set.

Add support for the setting, which can be used for example as:

```yaml
cluster:
  postgresql:
    extensions:
      - name: pg-oidc-validator
        image:
          reference: registry.codicelieve.com/pg-oidc-validator:v0.2-pg18-trixie
          pullPolicy: IfNotPresent
```


LoniasGR:
Hi! This is something that interests us, and seems pretty much completed. Could we get it merged? Thank you!

cfis:
I would also like to see the merged - thanks!

PR #855, opened 2026-04-26
chore: replace pg_cache_hit_ratio example with pg_database_size_bytes

philippemnoel:
## Summary
- The `pg_cache_hit_ratio` query used as the documented `customQueries` example (heap_blks_hit / heap_blks_read over `pg_statio_user_tables`) is a misleading default: it covers only user-table heap blocks and ignores index/catalog cache behavior, which is usually what matters for cache-pressure analysis.
- Its only live consumer — MCC's `mcc-customer` Cache Hit Ratio panel — has been switched (paradedb/mcc#105) to derive cache hit ratio from the standard `cnpg_pg_stat_database_blks_hit` / `cnpg_pg_stat_database_blks_read` metrics that CNPG exports by default. The companion BYOC change (paradedb/terraform-paradedb-byoc#946) drops the corresponding live customQueries entry on the MCC monitoring DB.
- Replace the `pg_cache_hit_ratio` example everywhere it appears in this chart with a neutral `pg_database_size_bytes` query, so the `customQueries` mechanism is still demonstrated and tested without implying a recommended (and misleading) metric.

Files touched:
- `charts/paradedb/values.yaml` — commented example block
- `charts/paradedb/examples/custom-queries.yaml`
- `charts/paradedb/test/monitoring/01-monitoring_cluster.yaml` + `-assert.yaml`

## Test plan
- [ ] `helm template` / chart test (`kuttl` or whatever runs the monitoring assert) still passes with the swapped query.


philippemnoel:
Opened against the wrong upstream by mistake — this should target paradedb/charts (a downstream fork). Closing. Sorry for the noise!

PR #890, opened 2026-06-01
feat: support namespaced webhook configurations for multi-operator deployments

maxlengdell:
Add config.namespacedWebhooks option that, when enabled:
- Appends the operator namespace to webhook configuration names
- Sets ENABLE_WEBHOOK_NAMESPACE_SUFFIX env var on the operator
- Adds namespaceSelector to webhook entries to scope them to the operator's namespace

This allows multiple CloudNativePG operators to run in single-namespace mode on the same cluster without webhook name collisions.

### DIFF
--- a/charts/cloudnative-pg/values.yaml
+++ b/charts/cloudnative-pg/values.yaml
@@ -78,6 +78,9 @@ config:
   # events across the entire Kubernetes cluster or if its focus should be
   # narrowed down to the specific namespace within which it has been deployed.
   clusterWide: true
+  # -- When set to true, appends the operator namespace to webhook configuration
+  # names to avoid collisions when running multiple operators in namespaced mode.
+  namespacedWebhooks: false
   # -- The content of the configmap/secret, see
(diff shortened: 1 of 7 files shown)


maxlengdell:
Should also update the ClusterRole and ClusterRoleBinding convention

charts/cloudnative-pg/values.yaml
```diff
@@ -78,27 +78,29 @@ config:
   # events across the entire Kubernetes cluster or if its focus should be
   # narrowed down to the specific namespace within which it has been deployed.
   clusterWide: true
+  # -- When set to true, appends the operator namespace to webhook configuration
+  # names to avoid collisions when running multiple operators in namespaced mode.
+  namespacedWebhooks: false
```
guille-work, line 83: Would it make more sense to fold this into `clusterWide` so that if `clusterWide` is false the webhooks are always namespaced?

I also wonder if this feature could allow for moving the webhooks permissions from cluster-wide to commonRules, allowing them to be namespace scoped as well, which would greatly remove the cluster-wide permissions needed to install the operator.

PR #928, opened 2026-06-24
feat: Adding support for the 'new' barman ObjectStore and ScheduledBackups

drew-viles:
We need to get a tagged release cadence going with this as the upstream chart is taking a little too long to move forward with supporting this stuff - I get they have a lot on, but we need these features now so this allows us to work in parallel with what they are working on. They have PRs open so there is no point raising this their and adding to their workload.

This is a merging of the `fixing-release-issues` branch that is currently running across all clusters, merged into the latest code form upstream.


drew-viles:
dang it, wrong one again - sorry

PR #924, opened 2026-06-20
feat( cluster ): plugins and barman cloud cnpg-i plugin support

itay-grudev:
Migrating from the Built-in CloudNativePG Backup to the Barman Cloud CNPG-I 
plugin is a straightforward, single step process, as the chart creates the 
necessary `ObjectStore` resource as a helm hook before the cluster is updated.

All you have to do is change the backup method from `barmanObjectStore` to 
`plugin` and specify the plugin name. The same change should also be applied to
your scheduled backups.

```diff
backups:
  enabled: true
- method: barmanObjectStore
+ method: plugin
+ pluginConfiguration:
+   name: barman-cloud.cloudnative-pg.io
  scheduledBackups:
    - name: daily-backup
      schedule: "0 0 0 * * *"
      backupOwnerReference: self
-     method: barmanObjectStore
+     method: plugin
+     pluginConfiguration:
+       name: barman-cloud.cloudnative-pg.io
```

closes: #830
closes: #910
closes: #838
closes: #775

PR #940, opened 2026-07-06
test(cluster): raise scheduledbackups cleanup timeout to 10m

aqeelat:
The `barman-plugin-scheduledbackups` test intermittently fails with `context deadline exceeded` during cleanup. The test logic passes; only namespace teardown exceeds the 5m budget.

**Root cause:** the CNPG instance pod has `terminationGracePeriodSeconds=1800` and graceful Postgres shutdown (bounded by `smartShutdownTimeout` ~180s) takes ~2–3m. On GitHub runners (kind-in-Docker on a constrained VM) that stretches past 5m under load. No stuck finalizer — Cluster and ObjectStore carry none (the plugin sets no ObjectStore finalizer).

Verified by reproducing on an isolated, well-resourced cluster: full cleanup completes in ~3m (pod teardown ~2m + namespace GC ~40s). The 5m budget is simply too tight for the slow-runner case.

**Fix:** raise `cleanup: 5m → 10m`. Shutdown is bounded by `smartShutdownTimeout`, so this will not grow unbounded; 10m gives safe headroom. Confirmed green in fork CI.

Supplements #924.

### DIFF
--- a/charts/cluster/test/barman-plugin-scheduledbackups/chainsaw-test.yaml
+++ b/charts/cluster/test/barman-plugin-scheduledbackups/chainsaw-test.yaml
@@ -7,7 +7,7 @@ spec:
   timeouts:
     apply: 1s
     assert: 2m
-    cleanup: 5m
+    cleanup: 10m
   steps:
     - name: Install the cluster with ScheduledBackups
       try:


aqeelat:
Closing in favor of a proper fix: set a low `spec.stopDelay` on the test cluster so the instance is force-stopped promptly at teardown (interrupting the still-running immediate backup) instead of waiting it out. The graceful shutdown is correct; it just doesn't need to complete a backup the test doesn't require. Will open a follow-up PR.

PR #943, opened 2026-07-08
[cluster]: don't fire HA alerts on replica clusters

Allex1:
On a replica cluster the designated primary runs a WAL receiver, lowering the computed standby count by one, so CNPGClusterHAWarning fired permanently. 
Gated both HA rules on `cnpg_collector_replica_mode == 0` so they only fire on primary clusters.


Allex1:
@itay-grudev @phisco May I get a CI workflow start and a review ?

charts/cluster/Chart.yaml
```diff
@@ -22,7 +22,7 @@ description: Deploys and manages a CloudNativePG cluster and its associated reso
 kubeVersion: ">=1.29.0-0"
 icon: https://raw.githubusercontent.com/cloudnative-pg/artwork/main/cloudnativepg-logo.svg
 type: application
-version: 0.8.1
+version: 0.8.2
```
itay-grudev: Could you please undo this change. We have a release procedure that's responsible for versioning. This isn't helping. The opposite actually. Please also undo the README change.

itay-grudev (changes requested):

charts/cluster/Chart.yaml
Allex1: done

Allex1:
@itay-grudev PTAL

Allex1:
@itay-grudev @phisco  PTAL

Allex1:
@itay-grudev  @phisco  PTAL

itay-grudev (approved):

itay-grudev (approved):

PR #944, opened 2026-07-09
feat(alerting): allow extra labels on PrometheusRules alerts

wanix:
draft only, to verify workflows


wanix:
sorry, miss-click on dest :(

PR #956, opened 2026-07-15
fix(cluster): consistent labels across all cluster chart resources

itay-grudev:
Fixes:
* Adds the `cnpg.io/cluster` label on resources created by the chart so they can be pin-pointed to the cluster they blong to.
* Applies `cluster.additionalLabels` to all resources
* Fixes: `cnpg.io/reload: ""` to `cnpg.io/reload: "true"`
* Adds `app.kubernetes.io/component` label to backups, poolers and the helm test

The end objective is to have consistent labels across all resources created by the chart.


phisco (approved):
🙏

Issue #561, opened 2025-04-17
OCI artifact tag gets updated on push to main


chrede88:
The newest release tag (`0.23.2`) gets updated on push to main via the [release-publish workflow](https://github.com/cloudnative-pg/charts/blob/main/.github/workflows/release-publish.yml). This is not great as the digest of the tag is also updated, which triggers redeployment of the chart for people running the cloudnative operator (even with the release pinned to the tag).
I don't know which CI workflows should run on pushes to main, but releasing charts shouldn't be done unless a new release is ready.

iSE-stefan-may:
Having the same problem. Something like this might work?

https://github.com/cloudnative-pg/charts/blob/main/.github/workflows/release-publish.yml#L77
```shell
          shopt -s nullglob
          for pkg in .cr-release-packages/*.tgz; do
            if [ -z "${pkg:-}" ]; then
              break
            fi
            
            file=${pkg##*/}
            name=${file%-*}
            version=${file%.*}
            version=${version##*-}

            # Check if version is already published
            if helm show chart oci://ghcr.io/${GITHUB_REPOSITORY}/${name} --version "$version" > /dev/null 2>&1; then
              continue
            fi

            helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
            cosign sign --yes ghcr.io/"${GITHUB_REPOSITORY}"/"${name}":"${version}"
          done
```

bootc:
For people following/finding this issue, I've opened a new issue to track this: #952.

PR #957, opened 2026-07-15
fix(ci): don't re-push unchanged charts to GHCR

### LINKED ISSUE #952: release-publish workflow re-pushes unchanged charts to GHCR on every push to main
The `release-publish` workflow triggers on every push to `main`, and its "Push charts to GHCR" step re-pushes and re-signs **every** packaged chart on every run, regardless of whether the chart's version actually changed. This means OCI tags in GHCR are effectively mutable: pulling by tag without pinning a digest can get you a different digest for the same version tag on essentially any merge to `main`.

This was previously reported in #561 ("OCI artifact tag gets updated on push to main"), which described the same root cause and included a proposed fix. That issue was auto-closed by the stale bot in May 2026 without getting fixed, despite multiple users confirming it was still happening. I can't reopen it, so filing this as a fresh issue with a current reproduction, and linking back for the prior discussion/history.

## Root cause

`.github/workflows/release-publish.yml` triggers on:

```yaml
on:
  push:
    branches:
      - main
```

It runs `chart-releaser` with `CR_SKIP_EXISTING: true`, which correctly skips creating a duplicate **GitHub Release** for a chart version that's already been released. However, `cr package` packages every chart in the repo unconditionally on every run — packaging isn't gated by skip-existing, only the release/upload step is.

The custom step that follows then loops over everything `cr package` produced and unconditionally pushes it to GHCR:

```yaml
- name: Push charts to GHCR
  env:
    COSIGN_EXPERIMENTAL: 1
  run: |
    shopt -s nullglob
    for pkg in .cr-release-packages/*.tgz; do
      if [ -z "${pkg:-}" ]; then
        break
      fi
      helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
      file=${pkg##*/}
      name=${file%-*}
      version=${file%.*}
      version=${version##*-}
      cosign sign --yes ghcr.io/"${GITHUB_REPOSITORY}"/"${name}":"${version}"
    done
```

There is no check here for whether that version tag has already been pushed to GHCR. So any push to `main` that repackages a chart — even with no `Chart.yaml` version bump — re-pushes it under the same OCI tag, producing a new digest under an unchanged tag.

## Reproduction

[Run 29255674287](https://github.com/cloudnative-pg/charts/actions/runs/29255674287), triggered by #945 landing on `main` (commit `dd86d35e`):

- The commit only touches `charts/plugin-barman-cloud/templates/rbac.yaml`, `values.yaml`, `values.schema.json` and `README.md`.
- `charts/plugin-barman-cloud/Chart.yaml` is **not** modified — no version bump.
- The push still triggered `release-publish`, which repackaged and re-pushed `plugin-barman-cloud` to GHCR.

## Impact

- Anyone consuming these charts via `oci://ghcr.io/cloudnative-pg/charts/...` by tag (rather than digest) can receive silently-changed content for what looks like the same released version.
- This is particularly problematic for anyone digest-pinning for supply-chain reasons: the digest they pinned against can stop matching what's published under that tag, as noted in #561 by @makkes.
- It also undermines the value of `cosign sign`/provenance for that tag, since the signed artifact for a given tag isn't stable over time.

## Suggested fix

As proposed in #561 by @iSE-stefan-may, check whether the version is already published to GHCR before pushing:

```bash
shopt -s nullglob
for pkg in .cr-release-packages/*.tgz; do
  if [ -z "${pkg:-}" ]; then
    break
  fi

  file=${pkg##*/}
  name=${file%-*}
  version=${file%.*}
  version=${version##*-}

  # Skip if this version is already published, so unrelated pushes to
  # main don't produce a new digest under an unchanged tag.
  if helm show chart oci://ghcr.io/${GITHUB_REPOSITORY}/${name} --version "$version" > /dev/null 2>&1; then
    continue
  fi

  helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
  cosign sign --yes ghcr.io/"${GITHUB_REPOSITORY}"/"${name}":"${version}"
done
```

Alternatively/additionally, restricting `release-publish` to only run when a chart's `Chart.yaml` version actually changes (e.g. via `paths-filter` or comparing against the previous commit) would avoid the wasted packaging/signing work entirely, not just the GHCR push.

Related: #561

bootc:
The `release-publish` workflow re-pushed and re-signed charts to GHCR on every push to `main`, which made the OCI tags effectively mutable. Pulling `oci://ghcr.io/cloudnative-pg/charts/...` by tag could return different content for the same version, digest pins stopped matching what was published under the tag, and the cosign signature for a tag was not stable over time.

This adds a check for whether a version is already published to GHCR, and skips it if so. The name/version parsing moves above the push, since the check needs it.

```bash
if helm show chart oci://ghcr.io/"${GITHUB_REPOSITORY}"/"${name}" --version "${version}" > /dev/null 2>&1; then
  echo "Skipping ${name} ${version}: already published to GHCR"
  continue
fi
```

The fix isn't introducing a new policy so much as restoring parity: `cr` already skips both the GitHub release and the repo index for a version that exists, and the GHCR push was the one publishing path with no equivalent guard. All three targets now agree.

Closes #952. Refs #561, which reported the same root cause and was closed by the stale bot without a fix.

### Root cause

The GHCR push is a custom step bolted on after `cr`, so it sits outside `cr`'s skip-existing logic entirely, and `helm push` has no skip-if-exists semantics — it just overwrites the tag.

One correction to the analysis in #952, from reading `chart-releaser-action` v1.7.0's `cr.sh`: `cr` does **not** package every chart in the repo unconditionally. `lookup_changed_charts` does `git diff --name-only <latest_tag> -- charts/`, so it packages charts with *any file* changed since the last tag. That is still broader than "the version was bumped", so the bug and the reproduction in the issue both stand — a commit touching `charts/plugin-barman-cloud/templates/rbac.yaml` with no `Chart.yaml` bump does get repackaged and re-pushed — but the mechanism is file-level, not unconditional.

### Why the gh-pages Helm repo was never affected

Worth recording, since it explains why this only ever bit the OCI path. In `chart-releaser` v1.7.0:

- `CR_SKIP_EXISTING` makes `CreateReleases` call `GetRelease` and `continue` past `CreateRelease` (`releaser.go:332-337`), so the `.tgz` asset attached to an existing release is never re-uploaded.
- `UpdateIndexFile` only calls `addToIndexFile` when `indexFile.Get(packageName, packageVersion)` returns an error, i.e. only when the entry is missing. For an already-published version nothing is added, `update` stays `false`, and it returns early with "Index did not change" before any write, commit or push.

So `cr` still repackages the chart into `.cr-release-packages/` on a no-bump push, but the classic path simply discards that local `.tgz`. Only the GHCR step consumed it.

### Note on `changed_charts`

The action exposes a `changed_charts` output whose description makes it look like the natural gate here. It isn't usable for this: it's populated from the same file-level diff against the last tag, so it lists charts that were repackaged without a version bump.

### Note on triggering from releases instead

#952 suggests gating the workflow on an actual version change, and triggering from the release/tag events would be the cleaner expression of that — `RELEASE.md` already describes that design ("A release `cloudnative-pg-vX.Y.Z` should be automatically created by an action, which will then trigger the release action").

That is not viable without a token change. `cr` creates the releases with `CR_TOKEN: secrets.GITHUB_TOKEN`, and events triggered by `GITHUB_TOKEN` [do not create new workflow runs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow) — the exceptions are `workflow_dispatch`, `repository_dispatch` and `pull_request`, not `release` or `push: tags`. An `on: release: published` workflow would silently never fire and charts would stop reaching GHCR altogether.

Making it work needs `CR_TOKEN` switched to a PAT (the repo already has `REPO_GHA_PAT`, used by `release-pr.yml`), which trades the current failure mode for publishing that breaks quietly if the PAT expires. That felt like a maintainer decision rather than something to bundle into this fix, so I've kept the change self-contained. Happy to follow up with the split if you'd prefer it.

The check here also has the property that it self-heals: if a GHCR push fails after the release is created, the next run retries it, whereas a purely event-driven trigger would miss it permanently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -80,10 +80,21 @@ jobs:
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
             file=${pkg##*/}
             name=${file%-*}
             version=${file%.*}
             version=${version##*-}
+
+            # cr packages any chart with files changed since the last tag, not
+            # just ones whose version was bumped, and CR_SKIP_EXISTING only gates
+            # the GitHub release. Skip versions already in GHCR, otherwise an
+            # unrelated edit republishes a chart under an unchanged tag with a
+            # new digest.
+            if helm show chart oci://ghcr.io/"${GITHUB_REPOSITORY}"/"${name}" --version "${version}" > /dev/null 2>&1; then
+              echo "Skipping ${name} ${version}: already published to GHCR"
+              continue
+            fi
+
+            helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
             cosign sign --yes ghcr.io/"${GITHUB_REPOSITORY}"/"${name}":"${version}"
           done
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - main
 
+# Two quick merges/pushes to main can cause a chart publish race condition
+concurrency:
+  group: chart-release
+  cancel-in-progress: false
+
 permissions: read-all
 
 jobs:
@@ -77,22 +82,25 @@ jobs:
         run: |
           shopt -s nullglob
           for pkg in .cr-release-packages/*.tgz; do
-            if [ -z "${pkg:-}" ]; then
-              break
-            fi
-            file=${pkg##*/}
-            name=${file%-*}
-            version=${file%.*}
-            version=${version##*-}
+            name=$(helm show chart "${pkg}" | yq -r '.name')
+            version=$(helm show chart "${pkg}" | yq -r '.version')
+            repo="${GITHUB_REPOSITORY}/${name}"
 
             # cr packages any chart with files changed since the last tag, not
             # just ones whose version was bumped, and CR_SKIP_EXISTING only gates
             # the GitHub release. Skip versions already in GHCR, otherwise an
             # unrelated edit republishes a chart under an unchanged tag with a
             # new digest.
-            if helm show chart oci://ghcr.io/"${GITHUB_REPOSITORY}"/"${name}" --version "${version}" > /dev/null 2>&1; then
-              echo "Skipping ${name} ${version}: already published to GHCR"
+            if output=$(helm show chart oci://ghcr.io/"${GITHUB_REPOSITORY}"/"${name}" \
+                --version "${version}" 2>&1); then
+              echo "::notice::Skipping ${name} ${version}: already published to GHCR"
               continue
+            elif grep -q "not found" <<< "${output}"; then
+              : # genuinely absent — fall through to push and sign
+            else
+              echo "::error::Cannot verify ${name} ${version} in GHCR; refusing to push blind"
+              echo "${output}"
+              exit 1
             fi
 
             helm push "${pkg}" oci://ghcr.io/"${GITHUB_REPOSITORY}"
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -84,15 +84,14 @@ jobs:
           for pkg in .cr-release-packages/*.tgz; do
             name=$(helm show chart "${pkg}" | yq -r '.name')
             version=$(helm show chart "${pkg}" | yq -r '.version')
-            repo="${GITHUB_REPOSITORY}/${name}"
+            repo="ghcr.io/${GITHUB_REPOSITORY}/${name}"
 
             # cr packages any chart with files changed since the last tag, not
             # just ones whose version was bumped, and CR_SKIP_EXISTING only gates
             # the GitHub release. Skip versions already in GHCR, otherwise an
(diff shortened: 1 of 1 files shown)


itay-grudev:
I found three more issues. 
1. Two merges shortly apart could trigger a race condition, so I added a concurrency group.
2. The `name` and `version` parameters would be parsed incorrectly if the SemVer contains dashes.
3. There was a corner case when `helm show chart` fails due to an error, other than `not found`. I want the workflow to exit with code `1` in that case.

I would appreciate it if you could validate the correctness of my changes (https://github.com/cloudnative-pg/charts/pull/957/changes/5dea284bc38a299052c78e4e7fab501fad17da0d) before I merge this.

bootc:
Your changes look correct, other than I pushed one more commit to use the `$repo` variable which makes things a tiny bit more readable. If the "not found" text ever changes then the pipeline will start failing to publish altogether, but it'll fail safe so I think that's reasonable.

itay-grudev (approved):

PR #617, opened 2025-07-03
feat(cluster): advanced recovery target variables in helm chart

zied-jt:
# Context

This PR take commits from #399 and add tests for each added new key in `value.yaml` file.

Commits from @sadath-12
- **support: backupID in helm chart**
- **fix: correct backupID field**
- **re-order and add other supported target recoveries**
- **add missing target params**

Commits for added tests:
- **feat(cluster): Test restore with backup id**
- **feat(cluster): Test restore with backup name**
- **feat(cluster): Test restore with xid**
- **feat(cluster): Test restore with lsn**
- **feat(cluster): Test restore with time**
- **feat(cluster): Test restore with exclusive**
- **feat(cluster): Test restore with immediate**

## Tests

Test are executed with [chainsaw](https://github.com/kyverno/chainsaw), example:

```bash
cd charts/cluster/test/
chainsaw test --test-file chainsaw-test-xid.yaml
```
## Related PRs
- #399 is directly linked as it cherry-peek @sadath-12 works
- #570 as could have implemented the same logic.

## Fixes

PR #959, opened 2026-07-20
feat(cluster): support named recovery target

philippemnoel:
## What

- Expose `recovery.pitrTarget.name` -> `recoveryTarget.targetName` and `recovery.pitrTarget.backupID` -> `recoveryTarget.backupID` on the `cluster` chart, alongside the existing `time` -> `targetTime`.
- Document them in `values.yaml`, `values.schema.json`, the README, and `docs/Recovery.md`.
- Extend the `postgresql-minio-backup-restore` chainsaw test to cover the named path.

## Why

`recovery.pitrTarget` only supported a wall-clock `time`. A named restore point (`pg_create_restore_point`) recovers to an exact WAL location, which is deterministic. CloudNativePG can pick the base backup to start replay from for a time target, but not for a name (or LSN/xid), so it rejects a named target with `recoveryTarget: Required value: BackupID is missing` unless `backupID` is given. `backupID` is exposed in the same change for that reason.

## Backwards compatibility

`recovery.pitrTarget.name` and `recovery.pitrTarget.backupID` default to `""`, and the template only emits a field when set. A `pitrTarget` with just `time` (or empty) renders identically to before.

## Tests

- `helm lint charts/cluster`.
- `helm template ... --show-only templates/cluster.yaml` renders `recoveryTarget.targetName`/`backupID` when set, and nothing when `pitrTarget` is empty.
- The `postgresql-minio-backup-restore` chainsaw test creates a `no-bad-table` restore point after the good write, then recovers from the object store with `pitrTarget.name` + `backupID` (read from the source backup's `.status.backupId`), asserting the pre-restore-point table is present and the post-restore-point table absent.

### DIFF
--- a/charts/cluster/README.md
+++ b/charts/cluster/README.md
@@ -319,7 +319,9 @@ Kubernetes: `>=1.29.0-0`
 | recovery.pgBaseBackup.source.sslRootCertSecret.key | string | `""` |  |
 | recovery.pgBaseBackup.source.sslRootCertSecret.name | string | `""` |  |
 | recovery.pgBaseBackup.source.username | string | `""` |  |
-| recovery.pitrTarget | object | `{"time":""}` | Point in time recovery target. Specify one of the following: |
+| recovery.pitrTarget | object | `{"backupID":"","name":"","time":""}` | Point in time recovery target. Specify at most one of `time` or `name`. |
+| recovery.pitrTarget.backupID | string | `""` | Backup to start replay from (a Backup's `.status.backupId`); required with `name`, not `time` |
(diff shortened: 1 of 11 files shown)


silanosa:
Thanks for this PR @philippemnoel, this is exactly what's missing for our team's workflow: reverting a local test database to a precise pre-test state via `pg_create_restore_point`, without the clock-skew caveats of time-based targets.

@itay-grudev @phisco any chance this could get a review? Happy to help test it against our setup if useful.

Related: #617 exposes the full set of recovery target options. Happy with either landing, but this smaller one already unblocks the `pg_create_restore_point` workflow for us.

PR #967, opened 2026-07-23
fix #963 accept null postgres UID GID

### LINKED ISSUE #963: cluster does not allow null PostgresUID/PostgresGID
# Issue

cluster forces a default `PostgresUID/PostgresGID = 26` (see https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/pods.go#L456). It is impossible to remove this default. One can set a value >= 0, but not null.

# Expected behavior

There should be a way to set the values such that any containers should be created without SCC.

# Why this matter

In non-root environment like OpenShift/OKD, with restricted-v2/v3, it is not possible to set UIDs. So a pod without SCC will inherit one with a forced `UID/GID/fsGroup` in a dynamic range depending on the namespace, so it cannot be hard-coded in `values.yaml`. Right now, it is not possible to deploy a cluster of cnpg inside such a cluster.

# Suggestion of fix
1. in go code
    1. Check uid/git at https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/pods.go#L464C18-L464C37
If null, we ignore these values, so `RunAsUser` `RunAsGroup` `FSGroup` are not set in the `defaultContext`
    2. Also check uid/git not null at https://github.com/cloudnative-pg/cloudnative-pg/blob/v1.30.0/pkg/specs/pods.go#L471

1. in helm cluster chart
    1. Make crd `postgresUID` `postgresGID` optional (so they can be null) and remove the default value of `26`
    2. To avoid regression with other environments, the default `values.yaml` should contains the default `postgresUID`, `postgresGID` set to `26`

There is a case of regression for people setting `cluster.useTimescaleDBDefaults` to `timescaledb`. Somehow, this requires UID/GID=1000. These people should write the UID/GID in their own `values.yaml`.

Please tell me if you have another suggestion. One alternative would be to create a new field like `openshiftmode = true` like in some other chart, but it is a bit uglier because it is in contradiction with non null UID/GID.

antoinetran:
Fix: https://github.com/cloudnative-pg/charts/issues/963

Description: accept `-1` as value for UID/GID, it has a special meaning: `runAsUser` `runAsGroup` `fsGroup` will not be defined.

### DIFF
--- a/charts/cloudnative-pg/templates/crds/crds.yaml
+++ b/charts/cloudnative-pg/templates/crds/crds.yaml
@@ -5118,13 +5118,13 @@ spec:
               postgresGID:
                 default: 26
                 description: The GID of the `postgres` user inside the image, defaults
-                  to `26`
+                  to `26`. `-1` is a special value meaning `runAsGroup` and `fsGroup` of `podSecurityContext` must not be defined.
                 format: int64
                 type: integer
               postgresUID:
                 default: 26
                 description: The UID of the `postgres` user inside the image, defaults
-                  to `26`
+                  to `26`. `-1` is a special value meaning `runAsUser` of `podSecurityContext` must not be defined.
                 format: int64
                 type: integer
               postgresql:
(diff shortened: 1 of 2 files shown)


itay-grudev:
If your change is merged in the operator feel free to re-open the issue. This ticket and the associated PR make no sense until then.

Issue #523, opened 2025-02-21
Make the smartShutdownTimeout flag assignable in helm


OmarGadd:
We have recently had the need to set the smartShutdownTimeout in the cluster spec to a different value than the default it was not possible to set it in the chart. We would like to request for that feature

Expected solution:

**values.yaml**
```
 .... 
cluster:
.....
    smartShutdownTimeout: 60

```

PR #972, opened 2026-07-24
feat(cluster): expose shutdown and failover timing knobs

### LINKED ISSUE #892: [Cluster chart] Add support for failoverDelay
The CNPG `Cluster` CRD supports [`spec.failoverDelay`](https://cloudnative-pg.io/documentation/current/cluster_conf/#failover) (integer seconds — delay before a primary is failed over once judged unhealthy), but the `cluster` chart does not expose it as a values key. Setting `cluster.failoverDelay` in values is silently dropped by the template.

This makes it impossible to use the chart for clusters that need to ride out brief instance-manager / kubelet hiccups without spurious failovers — which on storage without WAL archiving (eg. file-thinpool, local PV) can leave the old primary divergent with no `pg_rewind` archive to fall back to.

## Reproducer

```yaml
# values.yaml
cluster:
  failoverDelay: 30
```

```bash
helm template foo cloudnative-pg/cluster --version 0.6.1 -f values.yaml | grep failoverDelay
# (no output — chart never templates the field)
```

## Proposed shape

Same pattern as #834 / #867 — small wire-up.

**`values.yaml`** (under `cluster:`):
```yaml
  # -- Seconds to wait before triggering a failover after the primary is considered unhealthy.
  # See https://cloudnative-pg.io/documentation/current/cluster_conf/#failover
  failoverDelay: 0
```

**`templates/cluster.yaml`** (next to `primaryUpdateMethod`):
```yaml
  failoverDelay: {{ .Values.cluster.failoverDelay }}
```

Happy to open a PR if the direction looks right.

wittfabian:
## What

Expose four CNPG `Cluster` spec fields as chart values, completing the shutdown/failover timing family that `cluster.stopDelay` (#941) started:

| values key | maps to | default |
|---|---|---|
| `cluster.smartShutdownTimeout` | `spec.smartShutdownTimeout` | 180 |
| `cluster.switchoverDelay` | `spec.switchoverDelay` | 3600 |
| `cluster.startDelay` | `spec.startDelay` | 3600 |
| `cluster.failoverDelay` | `spec.failoverDelay` | 0 |

The template never rendered these, so setting them in `values.yaml` was silently dropped and they stayed pinned to the operator defaults.

## Why

On a brief API-server connectivity loss on the node hosting the primary, CNPG fails the primary over. With the default `smartShutdownTimeout: 180`, the old primary spends up to 180s in smart shutdown waiting for idle client connections to drain before the standby can be promoted — turning a ~1 minute network blip into a multi-minute read-write outage. Lowering `smartShutdownTimeout` (and raising `failoverDelay` to ride out very short partitions) is the intended CNPG knob, but with the chart not exposing it the only workaround is an out-of-band `kubectl patch` on the `Cluster` CR.

This closes the open request in #892 (`failoverDelay`) and also covers the remaining fields from #523 (`smartShutdownTimeout`, `startDelay`, `switchoverDelay`).

## How

Same wire-up as #941:
- `templates/cluster.yaml`: `{{- with }}` blocks, so a field is only rendered when set (`failoverDelay: 0` = operator default therefore stays unset).
- `values.yaml`: documented keys at their operator defaults.
- `values.schema.json`: regenerated with helm-schema `v0.23.4`.
- `README.md`: regenerated with helm-docs `v1.14.2`.

Verified with `helm lint` (0 failures) and `helm template`: fields render when set; `failoverDelay: 0` is correctly not emitted.

Closes #892

---

Disclosure (per the project [AI policy](https://github.com/cloudnative-pg/governance/blob/main/AI_POLICY.md)): this change was drafted with AI assistance — see the `Assisted-by:` commit trailer. I have reviewed and understand every line and take responsibility for it.

### DIFF
--- a/charts/cluster/README.md
+++ b/charts/cluster/README.md
@@ -202,6 +202,7 @@ Kubernetes: `>=1.29.0-0`
 | cluster.enableSuperuserAccess | bool | `true` | When this option is enabled, the operator will use the SuperuserSecret to update the postgres user password. If the secret is not present, the operator will automatically create one. When this option is disabled, the operator will ignore the SuperuserSecret content, delete it when automatically created, and then blank the password of the postgres user by setting it to NULL. |
 | cluster.env | list | `[]` | Env follows the Env format to pass environment variables to the pods created in the cluster |
 | cluster.envFrom | list | `[]` | EnvFrom follows the EnvFrom format to pass environment variables sources to the pods to be used by Env |
+| cluster.failoverDelay | int | `0` | The amount of time in seconds to wait before triggering a failover after the primary instance was detected to be unhealthy. Defaults to 0 (immediate failover) when unset, per the operator default. |
 | cluster.imageCatalogRef | object | `{}` | Reference to `ImageCatalog` of `ClusterImageCatalog`, if specified takes precedence over `cluster.imageName` |
 | cluster.imageName | string | `""` | Name of the container image, supporting both tags (<image>:<tag>) and digests for deterministic and repeatable deployments: <image>:<tag>@sha256:<digestValue> |
 | cluster.imagePullPolicy | string | `"IfNotPresent"` | Image pull policy. One of Always, Never or IfNotPresent. If not defined, it defaults to IfNotPresent. Cannot be updated. More info: https://kubernetes.io/docs/concepts/containers/images#updating-images |
@@ -242,10 +243,13 @@ Kubernetes: `>=1.29.0-0`
 | cluster.securityContext | object | `{}` | Configure Container Security Context. See: https://cloudnative-pg.io/documentation/preview/security/ |
 | cluster.serviceAccountTemplate | object | `{}` | Configure the metadata of the generated service account |
 | cluster.services | object | `{}` | Customization of service definitions. Please refer to https://cloudnative-pg.io/documentation/current/service_management/ |
+| cluster.smartShutdownTimeout | int | `180` | The time in seconds reserved for the smart shutdown of Postgres to complete before the operator requests a fast shutdown. Reserve enough time for the fast phase (that is: stopDelay - smartShutdownTimeout). Defaults to 180 when unset, per the operator default. |
+| cluster.startDelay | int | `3600` | The time in seconds that is allowed for a PostgreSQL instance to successfully start up. The startup probe failure threshold is derived from this value using the formula ceiling(startDelay / 10). Defaults to 3600 when unset, per the operator default. |
 | cluster.stopDelay | int | `1800` | The time in seconds that is allowed for the instance to wait for the shutdown to complete before being forcefully terminated. Also sets the pod's terminationGracePeriodSeconds. Defaults to 1800 (30m) when unset, per the operator default. |
 | cluster.storage.size | string | `"8Gi"` |  |
 | cluster.storage.storageClass | string | `""` |  |
 | cluster.superuserSecret | string | `""` |  |
+| cluster.switchoverDelay | int | `3600` | The time in seconds that is allowed for a primary instance to gracefully shut down during a switchover. Defaults to 3600 (1h) when unset, per the operator default. |
 | cluster.walStorage.enabled | bool | `false` |  |
 | cluster.walStorage.size | string | `"1Gi"` |  |
 | cluster.walStorage.storageClass | string | `""` |  |
--- a/charts/cluster/templates/cluster.yaml
+++ b/charts/cluster/templates/cluster.yaml
@@ -56,6 +56,18 @@ spec:
   {{- with .Values.cluster.stopDelay }}
   stopDelay: {{ . }}
   {{- end }}
+  {{- with .Values.cluster.smartShutdownTimeout }}
+  smartShutdownTimeout: {{ . }}
+  {{- end }}
+  {{- with .Values.cluster.switchoverDelay }}
+  switchoverDelay: {{ . }}
+  {{- end }}
+  {{- with .Values.cluster.startDelay }}
+  startDelay: {{ . }}
+  {{- end }}
+  {{- with .Values.cluster.failoverDelay }}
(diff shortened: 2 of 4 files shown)


juzu5:
Thanks for this — it covers exactly what we need. One issue: `{{- with .Values.cluster.smartShutdownTimeout }}` drops `0`, which is valid for this field (skip the smart phase) and differs from the operator default of 180, so `smartShutdownTimeout: 0` silently becomes 180. Suggest `{{- if hasKey .Values.cluster "smartShutdownTimeout" }}` for this key. Context: with the default we measured ~204 s without a writable primary on a graceful primary stop; being able to set ~30 s is the fix. (Drafted with AI assistance.)

wittfabian:
Good catch — you're right, and I've pushed a fix (04a5b9c).

Confirmed against the operator: `smartShutdownTimeout` is the only field in this timing family that is a pointer ([`cluster_types.go`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/api/v1/cluster_types.go)), and its getter returns the pointed-to value whenever it is set instead of falling back on zero:

```go
SmartShutdownTimeout *int32 `json:"smartShutdownTimeout,omitempty"`

func (cluster *Cluster) GetSmartShutdownTimeout() int32 {
	if cluster.Spec.SmartShutdownTimeout != nil {
		return *cluster.Spec.SmartShutdownTimeout
	}
	return 180
}
```

So `0` is a real setting rather than "unset": in `TryShuttingDownSmartFast` a `smartTimeout` of 0 skips the smart phase and requests a fast shutdown right away. The CRD carries no `minimum` on the field and there is no webhook check against it either. `{{- with }}` dropped the key and the operator then applied 180 — exactly the silent fallback you describe.

Switched to key presence, as suggested:

```gotemplate
{{- /* smartShutdownTimeout accepts 0 (skip the smart phase), which `with` would drop, so test for the key */}}
{{- if hasKey .Values.cluster "smartShutdownTimeout" }}
smartShutdownTimeout: {{ .Values.cluster.smartShutdownTimeout }}
{{- end }}
```

I deliberately left the other three on `{{- with }}`, because there a dropped `0` and the operator default are the same thing — those getters fall back on zero themselves:

| field | `0` rendered | key dropped |
|---|---|---|
| `startDelay` | `GetMaxStartDelay()` → 3600 | 3600 |
| `switchoverDelay` | `GetMaxSwitchoverDelay()` → 3600 | 3600 |
| `stopDelay` (#941) | `GetMaxStopDelay()` → 1800 | 1800 |
| `failoverDelay` | CRD default 0 | 0 |

Verified with `helm template`: `0`, `30` and the default `180` all render, and an explicit `null` leaves the key unset; `helm lint` is clean. `values.yaml` now documents the `0` case, and README/schema were regenerated with the pinned helm-docs `v1.14.2` and helm-schema `0.23.4`.

Your ~204 s fits the picture — 180 s of smart shutdown waiting for idle connections to drain, plus the rest of the promotion. One thing to keep in mind for a ~30 s target: the operator zeroes `smartShutdownTimeout` itself (with a warning) when `stopDelay <= smartShutdownTimeout`. Not a concern at the chart default of `stopDelay: 1800`, just don't lower `stopDelay` to the same order of magnitude.

Disclosure (per the project [AI policy](https://github.com/cloudnative-pg/governance/blob/main/AI_POLICY.md)): this follow-up was drafted with AI assistance — see the `Assisted-by:` commit trailer. I have reviewed and understand every line and take responsibility for it.

danishedb:
Hi @wittfabian If you can add the test for this as well to cover these new parameters. You can use this [Guide](https://github.com/cloudnative-pg/charts/blob/main/charts/cluster/TESTING.md) To test it locally when you add.

wittfabian:
Thanks @danishedb — tests added in d2267a0, and the branch is rebased on current `main`.

Following the guidance in TESTING.md, I extended `postgresql-cluster-configuration` rather than adding a separate test. Two parts:

**1. The four new fields in the existing non-default cluster**, plus `stopDelay` — that one landed in #941 and had no coverage either, so the timing family is now complete:

```yaml
stopDelay: 900
smartShutdownTimeout: 120
switchoverDelay: 60
startDelay: 120
failoverDelay: 30
```

**2. A dedicated step for `smartShutdownTimeout: 0`** (`04-zero_smart_shutdown_timeout.yaml`). It needs its own cluster because a cluster can only carry one value per field, and 0 is precisely the value the template used to lose — so asserting it is what makes this a regression guard rather than a restatement of the happy path.

Ran it locally per TESTING.md on a kind cluster with the operator installed from `charts/cloudnative-pg`:

```
--- PASS: chainsaw/postgresql-cluster-configuration (8.09s)
- Passed  tests 1
- Failed  tests 0
```

And to confirm the new step actually bites, I reverted `templates/cluster.yaml` to the previous `{{- with }}` guard and re-ran:

```
| Install a cluster with smartShutdownTimeout set to zero | ASSERT | ERROR |
=== ERROR
* spec.smartShutdownTimeout: Invalid value: 180: Expected value: 0
-  smartShutdownTimeout: 0
+  smartShutdownTimeout: 180
```

The key is dropped, the CRD default puts 180 back, and the assertion fails — which is exactly the silent fallback this PR is fixing. Restored the template and the suite passes again.

One interaction I checked while I was in there: `barman-plugin-scheduledbackups` sets `stopDelay: 30`, so `stopDelay <= smartShutdownTimeout` and the operator zeroes the smart timeout itself — the fast teardown that test relies on. That behaviour is unchanged by this PR, since the CRD default of 180 already applied there before the chart rendered the field.

I did not run the tests that need MinIO, cert-manager or the Prometheus CRDs — those are left to CI.

Disclosure (per the project [AI policy](https://github.com/cloudnative-pg/governance/blob/main/AI_POLICY.md)): drafted with AI assistance — see the `Assisted-by:` commit trailer. I have reviewed and understand every line and take responsibility for it.

danishedb (approved):

PR #991, opened 2026-08-27
fix(cluster):  align  plugin ObjectStore name to  barmanObjectName

### LINKED ISSUE #975: Fix ObjectStore name in the cluster chart
Please fix the name:

https://github.com/cloudnative-pg/charts/blob/b21ad2dac7992215b8572007adb7448e3c8cc5db/charts/cluster/templates/backup-objectstore.yaml#L5

to value from `.Values.cluster.plugins.parameters.barmanObjectName`

https://github.com/cloudnative-pg/charts/blob/b21ad2dac7992215b8572007adb7448e3c8cc5db/charts/cluster/templates/cluster.yaml#L173

danishedb:
The pr fixes the name  name: {{ include "cluster.fullname" . }}-backups  in barman plugin object store to match the cluster 
barmanObjectName,

closes #975 
Signed-off-by: danishedb <danish.khan@enterprisedb.com>

### DIFF
--- a/charts/cluster/templates/_helpers.tpl
+++ b/charts/cluster/templates/_helpers.tpl
@@ -243,3 +243,12 @@ Given a list of objects, returns the first item that matches the key value pairs
   {{- end -}}
   {{- $required -}}
 {{- end -}}
+
+{{/*
+Name of the plugin's barmanObject Nameparameter when set, otherwise defaults to "<fullname>-backups".
+*/}}
+{{- define "cluster.barmanObjectName" -}}
+  {{- $barmanPlugin := include "helpers.fetchKeyValue" (list .Values.cluster.plugins (list (list "name" "barman-cloud.cloudnative-pg.io"))) | fromYaml -}}
+  {{- $parameters := coalesce $barmanPlugin.parameters dict -}}
+  {{- $parameters.barmanObjectName | default (printf "%s-backups" (include "cluster.fullname" .)) -}}
+{{- end -}}
--- a/charts/cluster/templates/backup-objectstore.yaml
+++ b/charts/cluster/templates/backup-objectstore.yaml
@@ -2,7 +2,9 @@
 apiVersion: barmancloud.cnpg.io/v1
 kind: ObjectStore
 metadata:
-  name: {{ include "cluster.fullname" . }}-backups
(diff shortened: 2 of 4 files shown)


litaocdl (approved):

hh24k:
@danishedb the default-name expression now lives in both cluster.yaml and backup-objectstore.yaml,  Would you consider moving it into a helper so both files share one source of truth?

hh24k (approved):

danishedb:
I have performed following testing to ensure we donnot break anything on existing clusters.

Scenario 1 — defaults (no barmanObjectName set):

Tested: upgrading a running cluster to this PR has no impact — same ObjectStore name, no Cluster spec change, no pod restart, WAL archiving and backups uninterrupted.

Scenario 2 — user sets barmanObjectName (the bug this PR fixes):

Tested : on the released chart this config deadlocks the cluster entirely — the ObjectStore is created under the hardcoded <fullname>-backups name while the Cluster references the custom name, which never exists. After upgrading to this PR the referenced ObjectStore is created and the cluster bootstraps with working archiving and backups (fresh installs work the same way).

Scenario 3 — user sets barmanObjectName and pre-creates their own ObjectStore with that name:

Tested : the upgrade silently replaces the  ObjectStore, re-targeting WALs and cutting retention with no alert — so I Am proposing to add backups.objectStore.create (default true) which, verified by test, leaves the self-managed ObjectStore completely untouched when set to false.

I will push the change for this `backups.objectStore.create`

danishedb:
Added this value now `backups.objectStore.create`

gbartolini (approved):

gbartolini:
@itay-grudev thoughts?

itay-grudev (changes requested):
Wouldn't it make sense to also add this for the recovery object store?

See: https://github.com/cloudnative-pg/charts/blob/742cbd7fb5692c456ce993327fac8f39e348180b/charts/cluster/values.yaml#L54

I would also recommend adding a parameter to the macro for a default suffix (backups/recovery).

danishedb:
> Wouldn't it make sense to also add this for the recovery object store?
> 
> See:
> 
> https://github.com/cloudnative-pg/charts/blob/742cbd7fb5692c456ce993327fac8f39e348180b/charts/cluster/values.yaml#L54
> 
> I would also recommend adding a parameter to the macro for a suffix (backups/recovery).


I have added the changes @itay-grudev . Tests are passing in my fork https://github.com/danishedb/charts/actions/runs/36035162823/job/107753342152?pr=5

itay-grudev (approved):

